PR #5, opened 2024-12-26
Bagel refactor

### LINKED ISSUE #2: Add NatSpec comments
Add NatSpec comments to the contract. Follow the spec here https://docs.soliditylang.org/en/latest/natspec-format.html and use the https://github.com/BreadchainCoop/breadchain contracts as a reference.

### LINKED ISSUE #3: Functional V1 Implementation
- [ ] Implement all functionality in #1 
- [ ] Implement tests 
- [x] Implement deployment utils

bagelface:
- Fix tests
- Refactor core functionality, variables, errors, functions, etc.
- Add NatSpec comments
- Bump Solidity version


foundry.toml
```diff
@@ -9,7 +9,7 @@ multiline_func_header = 'params_first_multi'
 sort_imports = true
 
 [profile.default]
-solc_version = '0.8.23'
+solc_version = '0.8.28'
```
RonTuretzky, line 12: For my knowledge , what is the benefit of going 28 instead of 23 here? 

And should we change the upstream foundry template to reflect this?

src/contracts/SavingCircles.sol
```diff
@@ -0,0 +1,276 @@
...
+import {ReentrancyGuard} from '@openzeppelin/utils/ReentrancyGuard.sol';
+
+import {ISavingCircles} from '../interfaces/ISavingCircles.sol';
+
+/**
+ * @title Saving Circles
+ * @notice Simple implementation of a rotating savings and credit association (ROSCA) for ERC20 tokens
+ * @author Breadchain Collective
+ * @author @RonTuretzky
+ * @author @bagelface
```
RonTuretzky: on the YD contract everyone has their ENS, wdyt our convention should be for tagging authors?

src/contracts/SavingCircles.sol
```diff
@@ -0,0 +1,276 @@
...
+  function initialize(address _owner) external override initializer {
+    __Ownable_init_unchained(_owner);
+  }
+
+  /**
+   * @notice Create a new saving circle
+   * @param _circle A new saving circle
+   */
+  function create(Circle memory _circle) external override {
+    bytes32 _id = keccak256(abi.encodePacked(_circle.name));
```
RonTuretzky: I was thinking a lot about how maybe  the ID should be a hash of all the variables + timestamp? so there can be name collisions. Not sure if it's worth it or not.

src/contracts/SavingCircles.sol
```diff
@@ -0,0 +1,276 @@
...
+   */
+  function depositFor(bytes32 _id, uint256 _value, address _member) external override nonReentrant {
+    _deposit(_id, _value, _member);
+  }
+
+  /**
+   * @notice Make a withdrawal from a specified circle
+   * @param _id Identifier of the circle
+   */
+  function withdraw(bytes32 _id) external override nonReentrant {
```
RonTuretzky: As discussed, we might want to make a `withdrawFor` function in order to allow withdrawing for another member in the event they are unable (like the `depositFor` functionality)

src/contracts/SavingCircles.sol
```diff
@@ -0,0 +1,276 @@
...
+  /**
+   * @notice Decommission an existing saving circle
+   * @dev Returns all deposits to members
+   * @param _id Identifier of the circle
+   */
+  function decommission(bytes32 _id) external override {
+    Circle storage _circle = circles[_id];
+
+    if (_circle.owner != msg.sender) {
+      if (!isMember[_id][msg.sender]) revert NotMember();
```
RonTuretzky, line 117: what's the value of having access control here for the circle members as opposed to permissionless? 

My concern is that practically what will happen is we /chain coop will have to decommission these, not the members themselves...

src/contracts/SavingCircles.sol
```diff
@@ -0,0 +1,276 @@
...
+  function _deposit(bytes32 _id, uint256 _value, address _member) internal {
+    Circle memory _circle = circles[_id];
+
+    if (_isDecommissioned(_circle)) revert NotCommissioned();
+    if (!isMember[_id][_member]) revert NotMember();
+    if (
+      block.timestamp < circles[_id].circleStart
+        || block.timestamp >= circles[_id].circleStart + (circles[_id].depositInterval * (circles[_id].currentIndex + 1))
+        || block.timestamp >= circles[_id].circleStart + (circles[_id].depositInterval * circles[_id].maxDeposits)
+        || balances[_id][_member] + _value > circles[_id].depositAmount
```
RonTuretzky, line 255: I'm wondering if allowing flexibility in the deposit amount instead of only allowing the full amount contributes to ux or makes it more confusing.. thoughts @diterra-code ?

src/interfaces/ISavingCircles.sol
```diff
@@ -0,0 +1,50 @@
...
+  error InvalidDeposit();
+  error InvalidCircle();
+  error NotCommissioned();
+  error NotMember();
+  error NotDecommissionable();
+  error NotWithdrawable();
+  error TransferFailed();
+
+  // External functions (state-changing)
+  function initialize(address owner) external;
```
RonTuretzky, line 50: isn't the wonderland convention to put the natspec here and inheritdoc the contract?

RonTuretzky (commented):
Have not gone over tests yet, will follow up

src/contracts/SavingCircles.sol
bagelface: tbh i cant think of any reason to not just use serial ids (like you would do if you were just minting new NFTs) and return the value when a new circle is created

src/contracts/SavingCircles.sol
bagelface: i usually like to avoid "godmode" functions, but if we think we need it we can add it. it just gives the circle members complete control which is a good or bad thing depending on perspective

src/interfaces/ISavingCircles.sol
bagelface: oh idk, is it? seems sort of strange but if that's convention i can change it

foundry.toml
bagelface: well no reason in particular. it's just the latest version that has bug fixes, improved gas efficiency, some new features. it's been out for like 3 months at this point. im not married to it, just thought it made sense to do

src/contracts/SavingCircles.sol
bagelface: ya i personally like ENS, i think we should just put it however the author requests. do you want me to use your ENS for this?

src/contracts/SavingCircles.sol
bagelface: added

src/contracts/SavingCircles.sol
bagelface: i changed to use serial ids

src/contracts/SavingCircles.sol
bagelface: if you let me know which ENS to use for you i will change. i changed mine

src/interfaces/ISavingCircles.sol
bagelface: fwiw i personally dont really like that convention

PR #22, opened 2025-01-23
Remove decomission access control

### LINKED ISSUE #16: Remove Access control from `decomission`
Motivation: 
If #8 is to be developed, it makes sense to enable automatic decommissions, the current design would prevent automation because the executor would have to be the members or owner.

bagelface:


src/contracts/SavingCircles.sol
```diff
@@ -114,26 +114,24 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
 
-      bool hasIncompleteDeposits = false;
-      for (uint256 i = 0; i < _circle.members.length; i++) {
-        if (balances[_id][_circle.members[i]] < _circle.depositAmount) {
-          hasIncompleteDeposits = true;
-          break;
-        }
+    bool hasIncompleteDeposits = false;
+    for (uint256 i = 0; i < _circle.members.length; i++) {
+      if (balances[_id][_circle.members[i]] < _circle.depositAmount) {
```
RonTuretzky, line 129: could it be if I claim after the circle has completed, that it will zero the `balances[id][myaddress]` ?

bagelface: You can only decommission if there are incomplete deposits. Withdrawing will zero out all member balances. Can you elaborate on the question/concern?

RonTuretzky: I can't remember, anyways I think I was wrong

RonTuretzky (approved):

Issue #9, opened 2025-01-19
Gas Optimizations


daopunk:
- using multiple if-revert statements is cheaper than using one long if-revert statement that uses `&&` / `||`.
- cache mapping lookups as function scoped variables rather than looking up the same mapping value multiple times in the same scope
- for function params, use calldata instead of memory when the function is not modifying the param

gas ops for for-loops:
- `++i` is cheaper than `i++`
- after sol 0.8.0, can use `unchecked` to increment vars that have no risk of overflow
- cache the length in a var

example of most efficient solidity for-loop:
```
uint256 len = arr.length;
for (uint256 i; i < len) {
   unchecked { ++i }
   // logic
}
```

valeriooconte:
@bagelface I worked on this issue with @exo404 but we don't know if we have done all the optimizations

bagelface:
> [@bagelface](https://github.com/bagelface) I worked on this issue with [@exo404](https://github.com/exo404) but we don't know if we have done all the optimizations

ok will take a look at your PR

RonTuretzky:
looks closed to me going by sub issues

Issue #15, opened 2025-01-20
Improve mapping lookups


RonTuretzky:
- [ ]  function scoped variables

RonTuretzky:
[ref here](https://github.com/BreadchainCoop/saving-circles/commit/80fe4b7a0b536d0251a83d8576ac3d157e017812)

valeriooconte:
@bagelface I worked on this issue with @exo404

bagelface:
> [@bagelface](https://github.com/bagelface) I worked on this issue with [@exo404](https://github.com/exo404)

ok will take a look at your PR

RonTuretzky:
closed as already implemented in other issues

Issue #10, opened 2025-01-19
Use Modifiers


daopunk:
this check is used 5 times, can be a modifier to clean up logic, reduce lines of code, and add more readability:
` if (_isDecommissioned(_circle)) revert NotCommissioned();`

same here
`if (!isMember[_id][_member]) revert NotMember();`

tbsoc:
@bagelface it looks like we had someone actually work on this issue in their own fork. Would you be able to open and review PRs for this?

valeriooconte:
@bagelface I worked on this issue with @exo404

bagelface:
> [@bagelface](https://github.com/bagelface) I worked on this issue with [@exo404](https://github.com/exo404)

ok will take a look at your PR

bagelface:
https://github.com/BreadchainCoop/saving-circles/pull/32

PR #27, opened 2025-05-15
Improving mapping lookups, use modifiers, gas optimization

valeriooconte:


src/contracts/SavingCircles.sol
```diff
@@ -24,6 +24,22 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
   mapping(address member => uint256[] ids) public memberCircles;
   mapping(address token => bool status) public allowedTokens;
 
+  /**
+   * @dev Requires circle is commissioned by checking if an owner is set
+   */
+  modifier circleDecomissioned(uint256 _id) {
```
bagelface, line 30: imo modifier names are more readable when they describe the requirement for calling them. in the case of these two modifiers it would be `onlyCommissioned(uint256 _id)` and `onlyMember(uint256 _id)`

exo404: Okok we can fix that thank you

bagelface:
So the biggest issue with this PR at the moment is that we do not want to have the Broodfund changes included in it. We just want to close the issues https://github.com/BreadchainCoop/saving-circles/issues/9, https://github.com/BreadchainCoop/saving-circles/issues/10, and https://github.com/BreadchainCoop/saving-circles/issues/15.

My recommendation would be to create a new fork off the current version of the repo and recreate only the changes that are relevant to those PRs (no Broodfund stuff), and then create a new PR for me to review.

PR #28, opened 2025-05-16
Fixed: Use modifiers (Issue #10)

valeriooconte:


valeriooconte:
@bagelface we created a new fork and introduced our changes as you suggested as. Let us know whether we should create another different fork for Broodfonds or keep working on this one

PR #31, opened 2025-05-20
Issue #9, #10 and #15

valeriooconte:
@bagelface now authors are at the top in the contract natspec comments

PR #32, opened 2025-05-31
Issue #10

valeriooconte:
@bagelface We made another PR adding only modifiers. We also edited onlyMember as you suggested, including a new input for the address.


exo404:
@bagelface @RonTuretzky this is the gas report after adding only the modifiers:

image

exo404:
Adding modifiers seems to increase gas costs @bagelface

RonTuretzky:
image
 I see gas savings in the median case when compared to this 
 
 from your last pr [here](https://github.com/BreadchainCoop/saving-circles/pull/31#issuecomment-2905270250)

exo404:
So only Max and Min increased, but modifiers are the only thing we changed in the _deposit function. Do you think that doing gas reports every change we make (I mean changes aiming for gas optimization) will help to identify the problem?

bagelface (approved):
lgtm

PR #34, opened 2025-07-20
Enhanced frontend fetching utils

RonTuretzky:
Will make frontend fetching easier


bagelface (approved):
This adds a lot of "bloat" to the contract imo, but if it's needed by the frontend at the moment I'm ok with it I guess. I would personally prefer an approach were we get a frontend design spec and then build the "view" functions around that rather than just creating a few functions to just dump the entire state of the contract.

RonTuretzky:
> This adds a lot of "bloat" to the contract imo, but if it's needed by the frontend at the moment I'm ok with it I guess. I would personally prefer an approach were we get a frontend design spec and then build the "view" functions around that rather than just creating a few functions to just dump the entire state of the contract.

So we don't have a frontend spec but I vibecoded a functional one here
 https://saving-circles.vercel.app/

Off this figma https://www.figma.com/design/L0Y4SoFUohC3ABNbkXXp28/Saving-circle-V1?node-id=0-1&t=kyiqLZqbETiS2UN8-1

And that's how I got basically the "requirements" , but I can double back and make a spec for this np

RonTuretzky:
Also can decouple some things to add less bloat

src/interfaces/ISavingCircles.sol
```diff
@@ -24,6 +28,104 @@ interface ISavingCircles {
...
+    uint256 circleId;
+    Circle circleInfo;
+    uint256 userBalance;
+    bool isMember;
+    bool isOwner;
+    bool isCurrentWithdrawer;
+    bool canWithdraw;
+    uint256 nextWithdrawTime;
+    uint256 depositWindowEnd;
+    bool isExpired;
```
bagelface: keep bools together for better storage packing. also makes its slightly more readable imo

bagelface (approved):
ya looks good. it's still a lot of extra contract code for retrieving data. It may be better designed as an "extension" contract so that you can use the savings contract without all this in favor of an event/subgraph based approach, but for now i'm ok with this so we can move fast on it

RonTuretzky:
> ya looks good. it's still a lot of extra contract code for retrieving data. It may be better designed as an "extension" contract so that you can use the savings contract without all this in favor of an event/subgraph based approach, but for now i'm ok with this so we can move fast on it

I think I'll iterate on that then , maybe I'll just expose the raw view methods here and add a "SavingCircleStateRetriever" that is invited with it

bagelface:
> > ya looks good. it's still a lot of extra contract code for retrieving data. It may be better designed as an "extension" contract so that you can use the savings contract without all this in favor of an event/subgraph based approach, but for now i'm ok with this so we can move fast on it
> 
> I think I'll iterate on that then , maybe I'll just expose the raw view methods here and add a "SavingCircleStateRetriever" that is invited with it

Implemented it here actually https://github.com/BreadchainCoop/saving-circles/pull/35 .. called it "SavingCirclesLens" inspired by Compound's "CompoundLens" contract (https://github.com/compound-finance/compound-protocol/blob/master/contracts/Lens/CompoundLens.sol)

PR #35, opened 2025-07-24
Move extra view logic for frontend to a separate stateless "lens" contract

bagelface:
With this approach we get to keep the contract containing the core saving circles functionality minimal but also get the benefit of all the view functions you implemented.


RonTuretzky (changes requested):
ok to approve this but would prefer to rename the contract to "SavingCirclesStateRetriever" over "Lens" as I believe it is more descriptive

bagelface:
is "Viewer" ok? "StateRetriever" feels a little long-winded

RonTuretzky:
lgtm

RonTuretzky (approved):

RonTuretzky:
actually need to review the tests more in depth

src/contracts/SavingCirclesViewer.sol
```diff
@@ -0,0 +1,349 @@
...
+  function getUserFinancialSummary(address _user) external view override returns (UserFinancialSummary memory summary) {
+    uint256[] memory userCircleIds = _getAllUserCircleIds(_user);
+    return _getUserFinancialSummary(_user, userCircleIds);
+  }
+
+  /// @inheritdoc ISavingCirclesViewer
+  function getUserMembershipStatus(address _user) external view override returns (UserMembershipStatus memory status) {
+    uint256[] memory userCircleIds = _getAllUserCircleIds(_user);
+    return _getUserMembershipStatus(_user, userCircleIds);
+  }
```
RonTuretzky, line 80: we might want to delete these, pretty sure the frontend is going to just use the`getComprehensiveUserData` function

src/contracts/SavingCirclesViewer.sol
```diff
@@ -0,0 +1,349 @@
...
+    for (uint256 i = 0; i < _circleIds.length; i++) {
+      circleDataArray[i] = _getUserCircleData(_user, _circleIds[i]);
+    }
+    return circleDataArray;
+  }
+
+  /// @inheritdoc ISavingCirclesViewer
+  function getTotalBalance(address _member) external view override returns (uint256 _totalBalance) {
+    return _getTotalBalance(_member);
+  }
```
RonTuretzky, line 105: same as comment above

RonTuretzky (approved):

PR #42, opened 2025-08-09
feat: Automated ERC20 Allowance-Based Deposits and Batch Operations

RonTuretzky:
## Summary
- Implements automated deposit functions leveraging ERC20 allowances to enable programmatic deposits
- Adds view method to enumerate all accounts eligible for automated deposits
- Provides batch deposit functionality for account abstraction flows

### New Functions Added

1. **`depositIfAllowed(uint256 id, address member)`**
   - Automatically deposits funds for members with sufficient ERC20 allowance
   - Can be called by anyone (keepers, automation services)
   - Only deposits the remaining amount needed

2. **`getEligibleAddressesForDeposit()`**
   - Returns all circle IDs and member addresses eligible for automated deposits
   - Checks allowances and deposit windows across all active circles
   - Useful for automation services to identify processable deposits

3. **`batchDepositIfAllowed(uint256 id, address[] members)`**
   - Processes multiple deposits in a single transaction
   - Ideal for DAOs or automation services managing multiple accounts
   - Gracefully skips members with insufficient allowance or completed deposits

### Implementation Details

- Added `InsufficientAllowance` error for better error handling
- Functions validate deposit windows and circle status
- Supports partial deposits (completes remaining amount needed)
- Comprehensive test coverage with 12 new unit tests

## Test Plan
✅ All existing tests pass (48 total)
✅ Added 12 new unit tests covering:
- Sufficient/insufficient allowance scenarios
- Partial deposits
- Non-member attempts
- Multiple circles
- Deposit window validation
- Batch operations with mixed eligibility
- Decommissioned circle handling

## Related Issue
Closes #39

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #44, opened 2025-09-25
Delegated saving circles extension

bagelface:
An alternative version of https://github.com/BreadchainCoop/saving-circles/pull/42 implemented as an extension contract so that it's not a required feature. Renamed to be called "DelegatedSavingCircles", which I think is a more accurate description of the functionality.


bagelface:
There's something fucked up with the linter. It throws and error on that line locally if it's all on a single line. And remote CI throws error if it's not.

PR #45, opened 2025-09-25
Delegated saving circles extension

bagelface:
An alternative version of https://github.com/BreadchainCoop/saving-circles/pull/42 implemented as an extension contract so that it's not a required feature. Renamed to be called "DelegatedSavingCircles", which I think is a more accurate description of the functionality.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211594503026616


bagelface:
There's something fucked up with the linter. It throws and error on that line locally if it's all on a single line. And remote CI throws error if it's not.

EDIT: Worked around this by changing `getEligibleAddressesForDeposit` to `getAddressesForDeposit` so that line 67 of "src/interfaces/IDelegatedSavingCircles.sol" didn't exceed 120 characters.

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -0,0 +1,194 @@
...
+import {IERC20} from '@openzeppelin/token/ERC20/IERC20.sol';
+
+/**
+ * @title DelegatedSavingCircles
+ * @notice Extension contract for delegated deposits in SavingCircles
+ * @dev This contract enables delegated ERC20 allowance-based deposits and batch operations
+ */
+contract DelegatedSavingCircles is IDelegatedSavingCircles, SavingCircles {
+  /// @notice The main SavingCircles contract
+  ISavingCircles public immutable SAVING_CIRCLES;
```
RonTuretzky, line 16: we're both inheriting and have an internal saving circle 🤔?

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * @notice Constructor
+   * @param _savingCircles Address of the main SavingCircles contract
+   */
+  constructor(address _savingCircles) {
+    SAVING_CIRCLES = ISavingCircles(_savingCircles);
+  }
+
+  /// @inheritdoc IDelegatedSavingCircles
+  function setDelegatedDepositsEnabled(bool _enabled) external override {
```
RonTuretzky, line 30: we should allow an account abstracted entrypoint as well (sig)

bagelface: will make a new issue for this

src/contracts/DelegatedSavingCircles.sol
bagelface: fixed

src/contracts/DelegatedSavingCircles.sol
bagelface: https://github.com/BreadchainCoop/saving-circles/issues/50

RonTuretzky:
@bagelface what's the reasoning for making another contract instead of adding into the saving circle itself? something about this implementation pattern is throwing me off. alternative approach in #46

bagelface:
> @bagelface what's the reasoning for making another contract instead of adding into the saving circle itself? something about this implementation pattern is throwing me off. alternative approach in #46

It's an optional feature that won't necessarily be needed or wanted by everyone. The thought was that this pattern allows us to keep the saving circle implementation as simple as possible. I'm open to conversation about it, but I think it sort of depends on what the purpose of this repo is

RonTuretzky:
> Member

I guess I think everyone is going to want this , but I can see the perspective that it is optional. I'm worried that perhaps this will complicate things down the line but generally can't think of short term short comings other than that we need to create multiple deploy scripts. ok to merge this

RonTuretzky (approved):

RonTuretzky:
@bagelface be advised that some feedback is that this is really less of "delegated" desosits and more "automated" deposits as we already have the feature where you can deposit on someone's behalf, this is lettering the saving circle itself make the deposit on your behalf, but ONLY GIVEN an allowance

bagelface:
> @bagelface be advised that some feedback is that this is really less of "delegated" desosits and more "automated" deposits as we already have the feature where you can deposit on someone's behalf, this is lettering the saving circle itself make the deposit on your behalf, but ONLY GIVEN an allowance

Doesn't "automated" imply there is a mechanism that does the depositing automatically? These changes just add functionality to allow any address to deposit on another addresses behalf if allowed. I think "delegated" is appropriate. It only becomes automated if you set up some external mechanism to call the `depositIfAllowed` or `batchDepositIfAllowed` functions.

PR #52, opened 2025-10-20
feat: invite function porting from SafetyNet

### LINKED ISSUE #51: Invite function integration from Safety Net

exo404:


script/InviteGenerator.sol
```diff
@@ -0,0 +1,102 @@
...
+  constructor(string memory inviteSigningDomain, string memory inviteSignatureVersion, string memory structName) {
+    if (bytes(inviteSigningDomain).length == 0) revert InvalidSigningDomain();
+    if (bytes(inviteSignatureVersion).length == 0) revert InvalidSignatureVersion();
+    _inviteDomainNameHash = keccak256(bytes(inviteSigningDomain));
+    _inviteDomainVersionHash = keccak256(bytes(inviteSignatureVersion));
+    string memory inviteTypeString = string(abi.encodePacked('Invite(uint256 ', structName, 'Id,uint256 nonce)'));
+    _inviteTypeHash = keccak256(bytes(inviteTypeString));
+  }
+  /// @notice Returns the struct hash of an invite
+
```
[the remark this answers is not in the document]
exo404: [d257086](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/d2570860c7ad0000570dbab299bcaaa6eacc7149)

exo404:
@RonTuretzky @bagelface all tests are passing I think we can already merge this

script/InviteGenerator.sol
```diff
@@ -0,0 +1,102 @@
...
+  bytes32 private _inviteDomainNameHash;
+
+  /// @notice Hashed signing version used for EIP-712 signatures
+  bytes32 private _inviteDomainVersionHash;
+
+  /// @notice EIP-712 type hash for invite signatures
+  bytes32 private _inviteTypeHash;
+
+  /// @notice EIP-712 domain type hash
+  bytes32 private constant _EIP712_DOMAIN_TYPEHASH =
```
RonTuretzky: why not just have the bytes and the string in a comment

src/contracts/SavingCircles.sol
```diff
@@ -19,12 +20,32 @@ import {ISavingCircles} from 'interfaces/ISavingCircles.sol';
...
   uint256 public constant MINIMUM_MEMBERS = 2;
 
+  /// @notice Invite signing domain name used for EIP-712 signatures
+  string private constant _INVITE_SIGNING_DOMAIN = 'StacksInvite';
+
+  /// @notice Invite signing version used for EIP-712 signatures
+  string private constant _INVITE_SIGNATURE_VERSION = '1';
+
+  /// @notice EIP-712 type hash for invite signatures
+  bytes32 private constant _INVITE_TYPEHASH = keccak256('Invite(uint256 circleId,uint256 nonce)');
```
RonTuretzky: have bytes and string in comment

src/contracts/SavingCircles.sol
```diff
@@ -19,12 +20,32 @@ import {ISavingCircles} from 'interfaces/ISavingCircles.sol';
 contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
   uint256 public constant MINIMUM_MEMBERS = 2;
 
+  /// @notice Invite signing domain name used for EIP-712 signatures
+  string private constant _INVITE_SIGNING_DOMAIN = 'StacksInvite';
```
RonTuretzky: have bytes and string in comment

src/contracts/SavingCircles.sol
```diff
@@ -19,12 +20,32 @@ import {ISavingCircles} from 'interfaces/ISavingCircles.sol';
...
+
+  /// @notice Invite signing version used for EIP-712 signatures
+  string private constant _INVITE_SIGNATURE_VERSION = '1';
+
+  /// @notice EIP-712 type hash for invite signatures
+  bytes32 private constant _INVITE_TYPEHASH = keccak256('Invite(uint256 circleId,uint256 nonce)');
+
+  /// @notice EIP-712 domain type hash
+  bytes32 private constant _EIP712_DOMAIN_TYPEHASH =
```
RonTuretzky: have bytes and string in comment

src/interfaces/ISavingCircles.sol
```diff
@@ -24,6 +24,14 @@ interface ISavingCircles {
     uint256 maxDeposits;
   }
 
+  /// @notice Struct defining an invite to join a circle
+  /// @param circleId ID of the Circle
+  /// @param nonce Unique nonce for the invite
+  struct Invite {
```
RonTuretzky: seems like a redundant struct imo

RonTuretzky (changes requested):
please add frontend documentation and compat snippet

exo404:
> please add frontend documentation and compat snippet

I'm working on it in a different PR @RonTuretzky

src/interfaces/ISavingCircles.sol
exo404: why? I can take the circleId and nonce from other sources but this seems easier to read in the code

script/InviteGenerator.sol
exo404: Do you want to delete all the constants and put them in the comment? @RonTuretzky

src/interfaces/ISavingCircles.sol
RonTuretzky: because it's just two ints

script/InviteGenerator.sol
RonTuretzky: yes best practice whe hashing strings is to just store the hash and put in a comment above the hash what it is ,there's no benefit to storing the string in the contract

src/contracts/SavingCircles.sol
exo404: [37d1cb9](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/37d1cb9745c9bded71bbb5a567edb92c45c66ab5)

src/contracts/SavingCircles.sol
exo404: [37d1cb9](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/37d1cb9745c9bded71bbb5a567edb92c45c66ab5)

src/contracts/SavingCircles.sol
exo404: [37d1cb9](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/37d1cb9745c9bded71bbb5a567edb92c45c66ab5)

script/InviteGenerator.sol
exo404: [262870d](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/262870dad510e56932891d12cf47415518a839d9)

src/interfaces/ISavingCircles.sol
exo404: Yes I agree but this is the whole purpose of using EIP712 standard. It works with structs so we can extend the invite type in the future. I was deleting it but then I thought this

exo404: Ref [here](https://eips.ethereum.org/EIPS/eip-712)

RonTuretzky: can you reference this? also can you provide another reference that is doing what you are doing?  [this](https://github.com/Layr-Labs/eigenlayer-contracts/blob/57c37ae5829a1d2abb42fcae05774201ac3716fe/src/contracts/permissions/KeyRegistrar.sol#L344) impl seems to be different

RonTuretzky: @exo404 can you ref specifically the part in the spec supports changing the struct?

exo404: @RonTuretzky there's a signing version parameter

exo404: image

exo404: So I think that we can change the struct and so the domain version if we need it

RonTuretzky: @exo404 did you happen to research to existing invitation link solidity implementations?

exo404:
> please add frontend documentation and compat snippet

btw there's this example in the EIP712 spec [here](https://[eips.ethereum.org/assets/eip-712/Example.js](https://eips.ethereum.org/assets/eip-712/Example.js))

src/interfaces/ISavingCircles.sol
exo404: > can you reference this? also can you provide another reference that is doing what you are doing? [this](https://github.com/Layr-Labs/eigenlayer-contracts/blob/57c37ae5829a1d2abb42fcae05774201ac3716fe/src/contracts/permissions/KeyRegistrar.sol#L344) impl seems to be different

you have to look at their '_calculateSignableDigest' function, that's the same function as our _hashInvite

exo404: I believe that the structHash depends on what we are signing with EIP712, correct me if I'm wrong

exo404: > @exo404 did you happen to research to existing invitation link solidity implementations?

I simply tried to follow the standard as I said last meeting

RonTuretzky: I'm starting to think that was a mistake and I should have looked for a reference

exo404: @bagelface I need a review here

bagelface: I think the "struct Invite" definition here is redundant because we are using variables that are already defined in the contract. Instead, we can implement the hashing function like this and still have a valid 712 implementation:

```solidity
function _hashInvite(uint256 _circleId, uint256 _nonce) private view returns (bytes32) {
    bytes32 _inviteTypehash = keccak256('Invite(uint256 circleId,uint256 nonce)');
    bytes32 _structHash = keccak256(abi.encode(_inviteTypehash, _circleId, _nonce));
    bytes32 _eip712DomainTypehash =
      keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)');
    ...
  }
```

exo404: Ok perfect, I changed it deleting the struct. Test are still passing [34f2310](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/34f23100e8f86ce7b7817854ea9f6ec53b88a315)

exo404:
@RonTuretzky can you review here so I can merge this and finalize the frontend component directly on stacks?

RonTuretzky:
@exo404  i'm not merging this until i see the frontend componenet 🙃

script/InviteGenerator.sol
```diff
@@ -26,12 +26,12 @@ contract InviteGenerator is Script {
   /// @notice Error for empty invite signature version
   error InvalidSignatureVersion();
 
-  constructor(string memory inviteSigningDomain, string memory inviteSignatureVersion, string memory structName) {
+  constructor(string memory inviteSigningDomain, string memory inviteSignatureVersion) {
```
RonTuretzky: small nit is that i don't understand why you keep insisting to have the strings on chain 😵

exo404: I'm sorry I didn't understand the issue, I'm changing it asap

exo404:
@RonTuretzky [e894b0c](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/e894b0c4ae90c909cdcf4d6486622d7fdb86be7a) this for frontend components

script/InviteGenerator.sol
exo404: [0b7ee91](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/0b7ee9118f80ed2545501eee3323a98bfd579767) @RonTuretzky

exo404:
@bagelface I updated the code with 
- [34f2310](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/34f23100e8f86ce7b7817854ea9f6ec53b88a315)
- [0b7ee91](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/0b7ee9118f80ed2545501eee3323a98bfd579767)
- [e894b0c](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/e894b0c4ae90c909cdcf4d6486622d7fdb86be7a)

the last one is for frontend components

src/contracts/SavingCircles.sol
```diff
@@ -137,6 +139,27 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
     emit CircleDecommissioned(_id);
   }
 
+  /// @inheritdoc ISavingCircles
+  function redeemInvite(uint256 _id, uint256 _nonce, bytes calldata _signature) external override nonReentrant {
```
RonTuretzky, line 160: @exo404 currently we're not checking we've reached the max number of invites and we're also not checking that enough members have joined before circle start time

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
exo404: @RonTuretzky I don't see a maximum number of members, am I missing something? Also where do you want to check the minimum members constant? Do you want to check it inside the redeem function?

RonTuretzky: the circle struct has a members array. this inherently implies that there is a max member count, as we init the circle with the member array. however this pr breaks this functionality, because we no longer have ot initalize the members array. this breaks the whole contract becuase the previous pattern basically had an assured static member size count when it was initalized, and this one does not. I am unsure that this contract design supports dyanmically joining , as it was never intended to be used like this. does that make sense?

exo404: I didn't think about this but you are 100% right. I remember the contract assumed this from the start. Ho do we proceed? I can try to refactor the contract to include the dynamic members number

RonTuretzky: you can start with some tests that check what happens through this behavior to understand if we've broken the contract or not. the safest approach would be to remove the members area from the struct,add a counter that increments for each invite acceptance, and  add in the decomission check to see if enough members have joined so the circle is locked until everyone has joined

exo404:
@RonTuretzky @franrolotti the invite feature is implemented, also with @diterra-code request for multiple invites generation. I fixed all the tests. Let me know if there's something to change. While you review this I'm focusing on Gelato's issue

frontend/GenerateInviteForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+      setIsGenerating(true);
+      setError(undefined);
+      setSignature(undefined);
+      setDigest(undefined);
+
+      try {
+        const circleId = BigInt(circleIdInput);
+        const nonce = BigInt(nonceInput);
+        const verifyingContract = verifyingContractInput.trim();
+        const privateKey = privateKeyInput.trim();
```
RonTuretzky: we cannot take the users private key in the frontend as a raw value, here is a reference for how to sign using a privy wallet integration https://github.com/communetxyz/sharehouse-app/blob/e92297e71da19818ac65983ea0bd88d1948d817e/app/dashboard/invites/page.tsx#L75

RonTuretzky (changes requested):

frontend/GenerateInviteForm.tsx
RonTuretzky: please investigate if it's possible to authorize multiple wallet signatures with one click

src/contracts/SavingCircles.sol
```diff
@@ -25,6 +26,8 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
   mapping(uint256 id => mapping(address member => bool status)) public isMember;
   mapping(address member => uint256[] ids) public memberCircles;
   mapping(address token => bool status) public allowedTokens;
+  mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
```
RonTuretzky: Shouldn't this be per circle per user?

app/invites.tsx
```diff
@@ -0,0 +1,569 @@
+import { useCallback, useMemo, useState } from 'react';
+import type { ReactNode } from 'react';
+import type { TypedDataDomain, TypedDataField } from 'ethers';
+import { BrowserProvider, Contract } from 'ethers';
+
+const PRIVY_APP_ID = (import.meta as any).env?.VITE_PRIVY_APP_ID ?? 'privy-app-id';
+const DEFAULT_SAVING_CIRCLES_ADDRESS =
```
franrolotti: We should use a fixed app config for the contract address

app/invites.tsx
```diff
@@ -0,0 +1,569 @@
...
+  if (invite.used === false) return 'open';
+  return 'checking...';
+}
+
+function SavingCircleInvitesContent() {
+  const { ready: privyReady, authenticated, user, login, logout } = usePrivy();
+  const { wallets, ready: walletsReady } = useWallets();
+
+  const [contractAddress, setContractAddress] = useState(DEFAULT_SAVING_CIRCLES_ADDRESS);
+  const [circleId, setCircleId] = useState('');
```
franrolotti: like the commune flow, invites should be scoped to a specific circle. the generator of the invites has to be the owner of the circle. Might be nicer to have `circleId` come from route params / app state rather than user input

app/invites.tsx
```diff
@@ -0,0 +1,569 @@
...
+      return;
+    }
+    const invitesToCreate = Number.isFinite(inviteCount) && inviteCount > 0 ? inviteCount : 1;
+
+    setIsGenerating(true);
+    try {
+      const signer = await getSignerFromPrivyWallet(activeWallet);
+      const network = await signer.provider?.getNetwork();
+      const chainId = network?.chainId ?? 0n;
+      const contract = new Contract(contractAddress, savingCirclesAbi, signer);
```
franrolotti: As we want only the circle owner to create invite links (like the commune invite flow). Since the ABI already exposes `owner()`, could we add an explicit owner check here?

And ideally add a “You’re not the owner” UI state instead of just a generic status.

app/invites.tsx
```diff
@@ -0,0 +1,569 @@
...
+      const baseUrl =
+        typeof window === 'undefined' ? 'https://savingcircles.local/invite' : `${window.location.origin}/invite`;
+
+      const generated: InviteLink[] = [];
+      let candidate = BigInt(Date.now());
+      while (generated.length < invitesToCreate) {
+        const alreadyUsed = await contract.usedNonces(parsedCircleId, candidate);
+        if (!alreadyUsed) {
+          const typedData = buildInviteTypedData(parsedCircleId, candidate, chainId, contractAddress);
+          const signature = await signInvite(signer, typedData);
```
franrolotti: This option doesn't batch all invites into a single signing step, nor hides the multiple signatures as we spoke in the meeting. We should solve this here

app/invites.tsx
```diff
@@ -0,0 +1,569 @@
...
+          }}
+        >
+          Refresh invite status
+        </button>
+      </div>
+
+      <section style={{ marginBottom: 22 }}>
+        <h3 style={{ margin: '0 0 8px 0' }}>Redeem an invite</h3>
+        <p style={{ margin: '0 0 10px 0', color: '#475569' }}>
+          Paste a signed invite to redeem it for the connected Privy wallet. This calls SavingCircles.redeemInvite.
```
franrolotti, line 479: May be better for the user to click an invite URL and the page auto-redeems those params for them (with a confirmation UI), without them having to copy/paste anything.

franrolotti (commented):
Right now this page feels more like an internal dev/admin tool than a product-facing “invite to a circle” flow frontend. I think the goal is to be closer to the commune invite UX.

We should refactor to keep the to keep all the EIP-712 + `usedNonces` logic but wrap it in a polished, app-styled UI like the commune invite page. 

Also we should refactor to

* Only circle owner can create invites.
* The page should be scoped to a specific circle where users don't type `circleId` or `contractAddress`
* Right now, multiple invites trigger multiple signature popups
* Use design system for the UI


My suggestion is to keep all the strong crypto/contract work from this PR, but refactor the component so this is the final Saving Circles invite experience (owner-only, single-circle, one-click create + URL-based redeem, commune-style UI), rather than a generic admin-style tool.

@RonTuretzky wdyt

RonTuretzky:
> Right now this page feels more like an internal dev/admin tool than a product-facing “invite to a circle” flow frontend. I think the goal is to be closer to the commune invite UX.
> 
> We should refactor to keep the to keep all the EIP-712 + `usedNonces` logic but wrap it in a polished, app-styled UI like the commune invite page.
> 
> Also we should refactor to
> 
> * Only circle owner can create invites.
> * The page should be scoped to a specific circle where users don't type `circleId` or `contractAddress`
> * Right now, multiple invites trigger multiple signature popups
> * Use design system for the UI
> 
> My suggestion is to keep all the strong crypto/contract work from this PR, but refactor the component so this is the final Saving Circles invite experience (owner-only, single-circle, one-click create + URL-based redeem, commune-style UI), rather than a generic admin-style tool.
> 
> @RonTuretzky wdyt

It should not be product facing, its just a reference for the frontend ppl to be able to have a working reference so a really janky page is fine as long as it is functional

exo404:
[5317da8](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/5317da8b21bea9e3746d7307dca46cb5d5d56d3a) fixes #64

exo404:
[1d466d5](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/1d466d5cb03cae2546e212de57c85647248955f5) fixes ##66

src/contracts/SavingCircles.sol
```diff
@@ -94,7 +87,25 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
     Circle storage _circle = circles[_id];
     if (isActive[_id]) revert AlreadyActive();
     if (msg.sender != _circle.owner) revert NotOwner();
-    if (_circle.members.length < MINIMUM_MEMBERS) revert InvalidMemberCount();
+    uint256 membersCount = 0;
+    for (uint256 i = 0; i < _circle.members.length; i++) {
```
RonTuretzky: do we really need two loops here or am I missing something?

src/contracts/SavingCircles.sol
```diff
@@ -198,26 +262,28 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    if (!isActive[_id]) revert NotActive();
     Circle memory _circle = circles[_id];
 
     return _circle.members[_circle.currentIndex];
   }
 
   /**
    * @dev Make a withdrawal from a specified circle
-   *      Permissionless: anyone can trigger the payout for the member whose turn it is to withdraw
+   *      A withdrawal must be made by a member of the circle, even if it is for another member.
```
RonTuretzky, line 269: when did we decide on this change?

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
exo404: I'm not sure I made this change tbh, but I can revert this. I see there's only one check in the function for this

src/contracts/SavingCircles.sol
exo404: I think we need to know the members array dimension before declaring it right? I doesn't work with dynamic arrays

src/contracts/SavingCircles.sol
exo404: [72f0c2c](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/72f0c2cac3f6bc2b4b1363cb599b577efd30d1c3)

src/contracts/SavingCircles.sol
```diff
@@ -291,9 +291,9 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
       balances[_id][_circle.members[i]] = 0;
     }
 
-    _circle.currentIndex = (_circle.currentIndex + 1) % _circle.members.length;
-    bool success = IERC20(_circle.token).transfer(_member, _withdrawAmount);
+    bool success = IERC20(_circle.token).transfer(_circle.members[_circle.currentIndex], _withdrawAmount);
```
RonTuretzky: i don't understand this change , why is this in this function? please decouple it so that this is determined in the func that is calling it (so withdraw for should determine whether the perosn is eligible for a withdrawl. also , i think it could be they're not the current index but can still withdraw? unsure?

src/contracts/SavingCircles.sol
RonTuretzky: https://github.com/BreadchainCoop/saving-circles/pull/70

src/contracts/SavingCircles.sol
```diff
@@ -63,48 +66,84 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+      if (member == address(0)) {
+        continue;
+      }
+      members[memberIndex] = member;
+      memberIndex++;
+    }
+    circles[_id].members = members;
 
-    emit CircleCreated(_id, _circle.members, _circle.token, _circle.depositAmount, _circle.depositInterval);
+    _circle.circleStart = block.timestamp;
```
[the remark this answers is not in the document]
exo404: we want to override it because we need to calculate the correct indexes

src/contracts/SavingCircles.sol
```diff
@@ -63,48 +66,84 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    uint256 len = _circle.members.length;
+    uint256 maxDelta = type(uint256).max - _circle.circleStart;
+    if (_circle.depositInterval > maxDelta / len) revert InvalidDepositInterval();
+    _circle.circleEnd = _circle.circleStart + (_circle.depositInterval * len);
 
-    return _id;
+    isActive[_id] = true;
+    emit CircleStarted(_id);
   }
```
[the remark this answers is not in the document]
RonTuretzky: @exo404 i agree

src/contracts/SavingCircles.sol
```diff
@@ -63,48 +66,84 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    uint256 memberIndex = 0;
+    for (uint256 i = 0; i < _circle.members.length; i++) {
+      address member = _circle.members[i];
+      if (member == address(0)) {
+        continue;
+      }
+      members[memberIndex] = member;
+      memberIndex++;
+    }
+    circles[_id].members = members;
```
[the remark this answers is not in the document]
exo404: I think I need to delete the check in the invite function

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: i think there's a difference between "circleMinimumStart" and "circleStart" which is basically what we're missing here otherwise its confusing.

src/contracts/SavingCircles.sol
```diff
@@ -133,10 +172,33 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+
+    if (_signer != _circle.owner) revert InvalidSigner();
+
+    usedNonces[_id][_nonce] = true;
+    isMember[_id][msg.sender] = true;
+    memberCircles[msg.sender].push(_id);
+    _circle.members.push(msg.sender);
+
+    emit InviteRedeemed(_id, msg.sender);
+  }
```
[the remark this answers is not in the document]
RonTuretzky: @exo404 wdyt?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: Because the frontend will pass circle struct with less members than the actual array lenght

src/contracts/SavingCircles.sol
```diff
@@ -211,23 +277,23 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
       balances[_id][_circle.members[i]] = 0;
     }
 
-    _circle.currentIndex = (_circle.currentIndex + 1) % _circle.members.length;
-    bool success = IERC20(_circle.token).transfer(_member, _withdrawAmount);
+    bool success = IERC20(_circle.token).transfer(_circle.members[_circle.currentIndex], _withdrawAmount);
     if (!success) revert TransferFailed();
+    _circle.currentIndex = (_circle.currentIndex + 1) % _circle.members.length;
 
     emit FundsWithdrawn(_id, _member, _withdrawAmount);
```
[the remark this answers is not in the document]
RonTuretzky: @exo404 is this addressed by your revert?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: yes me too, I'll add tests for this

src/contracts/SavingCircles.sol
```diff
@@ -211,23 +277,23 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
   /**
    * @dev Make a withdrawal from a specified circle
    *      Permissionless: anyone can trigger the payout for the member whose turn it is to withdraw
+   *      A withdrawal must be made by a member of the circle, even if it is for another member.
    */
-  function _withdraw(uint256 _id, address _member) internal {
+  function _withdraw(uint256 _id, address _member) internal onlyMember(_id, msg.sender) {
```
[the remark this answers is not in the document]
exo404: [60abb0d](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/60abb0d6eceab448f8f113c8fb8cba982e7d7a0b)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: what do you mean?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: Yes I implemented tests before but I was using the old contract to check it. It was ok with that, but maybe we need new tests for it? Let me know I got no problems doing new tests

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: [60abb0d](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/60abb0d6eceab448f8f113c8fb8cba982e7d7a0b)

exo404:
[check] what happens if no one withdraws?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: @RonTuretzky  wdyt?

src/contracts/SavingCircles.sol
```diff
@@ -63,48 +66,84 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
-      if (_member == address(0)) revert InvalidMemberAddress();
-      isMember[_id][_member] = true;
-      memberCircles[_member].push(_id);
+      address member = _circle.members[i];
+      if (member == address(0)) revert InvalidMemberAddress();
+      if (isMember[_id][member]) revert AlreadyMember();
+
+      isMember[_id][member] = true;
+      memberCircles[member].push(_id);
     }
```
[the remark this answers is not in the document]
exo404: [2824d76](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/2824d76feda6f47350961fc0ed05e3552e81bad4)

src/contracts/SavingCircles.sol
exo404: I leaved a comment there @RonTuretzky

src/contracts/SavingCircles.sol
exo404: [60abb0d](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/60abb0d6eceab448f8f113c8fb8cba982e7d7a0b)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: I don't understand please rewrite more clearly , which check, when is the frontend passing in what, how is this related to the comment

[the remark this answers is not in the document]
RonTuretzky: I don't understand please rewrite more clearly , which check, when is the frontend passing in what, how is this related to the comment

src/contracts/SavingCircles.sol
```diff
@@ -295,4 +361,24 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    bytes32 _structHash = keccak256(abi.encode(_inviteTypehash, _id, _nonce));
+    bytes32 _eip712DomainTypehash = 0x8b73c3c69bb8fe3d512ecc4cf759cc79239f7b179b0ffacaa9a75d522b39400f;
+    bytes32 _inviteDomainNameHash = 0xf50d3e48fa87e894899f86eba14c57c836bc6ffddd68251a158269ffdadc0cb1;
+    bytes32 _inviteDomainVersionHash = 0xc89efdaa54c0f20c7adf612882df0950f5a951637e0307cdcb4c672f298b8bc6;
+
+    bytes32 _domainSeparator = keccak256(
+      abi.encode(_eip712DomainTypehash, _inviteDomainNameHash, _inviteDomainVersionHash, block.chainid, address(this))
+    );
+
+    return keccak256(abi.encodePacked('\x19\x01', _domainSeparator, _structHash));
```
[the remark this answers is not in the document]
RonTuretzky: I think we can ignore this but let's make a backlog issue @brendaasfjnso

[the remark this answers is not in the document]
RonTuretzky: I think we can ignore this but let's make a backlog issue @brendaasfjnso

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: There is a difference between "don't start before this time" and "this started at this time"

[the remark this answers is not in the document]
RonTuretzky: There is a difference between "don't start before this time" and "this started at this time"

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: Yes we need good coverage

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: ok I see but we have got only the second one. Also does make sense to have a "don't start before this time" variable?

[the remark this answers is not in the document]
RonTuretzky: What are we initializing then in the create? Which one is it supposed to be? If it's not supposed to be "don't start before this" then the create should not be initializing it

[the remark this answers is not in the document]
RonTuretzky: What are we initializing then in the create? Which one is it supposed to be? If it's not supposed to be "don't start before this" then the create should not be initializing it

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: Ok so first of all, I don't think ordering is important because we are not considering it. So the first part of the Copilot comment is useless. The second thing is that currently we pass a members array with the maximum size specified in the frontend, but without all the members addresses. So the circle passed to the create function got a members array with some address(0) in it. Maybe we can change the struct to take a new field called size to avoid this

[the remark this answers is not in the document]
exo404: @RonTuretzky

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: We actually don't touch the circleStart field in the create function, but we are passing the circle struct containing a circleStart value, that's it

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: [f75f632](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/f75f632c5a2792fca3fb1916779c11901495b584)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: [436855f](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/436855ff7515b32b7e5e5208c477fe3a30ea876b)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
exo404: [3e0a9ef](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/3e0a9efef701a7eb86e3837ce5f9b9230673f369)

src/contracts/SavingCircles.sol
```diff
@@ -63,48 +66,81 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+      if (member != address(0)) {
+        tempMembers[membersCount] = member;
+        membersCount++;
+      }
+    }
+    if (membersCount < MINIMUM_MEMBERS) revert InvalidMemberCount();
+
+    address[] memory members = new address[](membersCount);
+    for (uint256 i = 0; i < membersCount; i++) {
+      members[i] = tempMembers[i];
```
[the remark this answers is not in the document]
exo404: This was my old idea but I think we resolved this with Ron's PR

exo404:
> [check] what happens if no one withdraws?

If no one claims the circle, no one can deposit because the claim index can't be updated but with the withdraw function. So the circle is freezed unless it expires or someone use the withdraw function @RonTuretzky 

lines 318-322 stop anyone to deposit until currentIndex is updated

src/contracts/SavingCircles.sol
```diff
@@ -71,9 +71,8 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
 
     for (uint256 i = 0; i < _circle.members.length; i++) {
       address member = _circle.members[i];
-      if (member == address(0)) revert InvalidMemberAddress();
+      if (member == address(0)) continue;
```
RonTuretzky: wait why did we delete teh revert here?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: ok this is what is confusing copilot , and its correct in the sense that the naming is ambigious. lets rename it to "effectiveCircleStartTime" which is only set when the circle actually starts, and make sure during circle creation its 0. we can also use this instead of the isActive mapping, but I kind of want to push this PR through and not refactor anyfurther

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: why are we even doing this during the start function? why don't we update the circle member count and the members array during the redeem invite function? whats the utility of having it in this function?

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
exo404: the problem with the members list is that now we got a dynamic list updated everytime we redeem a function, so a static array in the circle struct is hard to manage. I'm trying to refactor this using a mapping instead of the static array

exo404:
@RonTuretzky @franrolotti I added a circleMembers mapping to replace the members array. I renamed the circleStart to effectiveCircleStartTime and I added a revert if it's not set to zero at circle creation. Now we update the membersMapping everytime we redeem an invite. I updated the Viewer and Delegated contract to match these changes and fixed the tests.

RonTuretzky:
> @RonTuretzky @franrolotti I added a circleMembers mapping to replace the members array. I renamed the circleStart to effectiveCircleStartTime and I added a revert if it's not set to zero at circle creation. Now we update the membersMapping everytime we redeem an invite. I updated the Viewer and Delegated contract to match these changes and fixed the tests. 

Fuck another massive change... Will review tmrw

RonTuretzky:
> @RonTuretzky @franrolotti I added a circleMembers mapping to replace the members array. I renamed the circleStart to effectiveCircleStartTime and I added a revert if it's not set to zero at circle creation. Now we update the membersMapping everytime we redeem an invite. I updated the Viewer and Delegated contract to match these changes and fixed the tests. 

Fuck another massive change... Will review tmrw

exo404:
> > @RonTuretzky @franrolotti I added a circleMembers mapping to replace the members array. I renamed the circleStart to effectiveCircleStartTime and I added a revert if it's not set to zero at circle creation. Now we update the membersMapping everytime we redeem an invite. I updated the Viewer and Delegated contract to match these changes and fixed the tests.
> 
> Fuck another massive change... Will review tmrw

this was the only option to fix this technical debt. Let me know if it's all clear

test/unit/SavingCirclesUnit.t.sol
```diff
@@ -64,19 +74,111 @@ contract SavingCirclesUnit is Test {
...
+    bytes32 domainSeparator = keccak256(
+      abi.encode(
+        eip712DomainTypehash, inviteDomainNameHash, inviteDomainVersionHash, block.chainid, address(savingCircles)
+      )
+    );
+
+    bytes32 digest = keccak256(abi.encodePacked('\x19\x01', domainSeparator, structHash));
+    (uint8 v, bytes32 r, bytes32 s) = vm.sign(_signerKey, digest);
+    return abi.encodePacked(r, s, v);
   }
```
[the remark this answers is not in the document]
RonTuretzky: @exo404 seems important

src/contracts/SavingCircles.sol
```diff
@@ -191,7 +178,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
     usedNonces[_id][_nonce] = true;
     isMember[_id][msg.sender] = true;
     memberCircles[msg.sender].push(_id);
-    _circle.members.push(msg.sender);
+    circleMembers[_id].push(msg.sender);
```
RonTuretzky, line 178: I dont see a max count validation here? I guess we are just trusting that the owner issues a finite amount of members but maybe we should add a comment to this

RonTuretzky (changes requested):
recent changes look good, small nit and copilot comments

app/invites.tsx
```diff
@@ -0,0 +1,606 @@
...
+export const INVITE_TYPEHASH = '0xd86e498a74dbfe863d870d4811dddab9c7f3922d6c0d6656504984bd9a8607a3';
+export const INVITE_DOMAIN_NAME = 'StacksInvite';
+export const INVITE_DOMAIN_VERSION = '1';
+
+const savingCirclesAbi = [
+  'function usedNonces(uint256 id, uint256 nonce) view returns (bool)',
+  'function redeemInvite(uint256 id, uint256 nonce, bytes signature)',
+  'function isMember(uint256 id, address member) view returns (bool)',
+  'function owner() view returns (address)',
+  'function getCircle(uint256 id) view returns (address owner, address[] members, uint256 currentIndex, uint256 depositAmount, address token, uint256 depositInterval, uint256 circleStart, uint256 circleEnd)'
```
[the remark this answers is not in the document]
RonTuretzky: @exo404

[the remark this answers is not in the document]
exo404: [f0fee14](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/f0fee142b2f8543d7292d2d0438082e0e2c5b281)

test/unit/SavingCirclesUnit.t.sol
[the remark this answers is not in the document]
exo404: [5201e5f](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/5201e5fbf12872dd19f8a2d163058a14d88b8569)

[bf432fe](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/bf432fe579f5f5734fcebbd813e1cc604b1bf2ea)

[ed893b6](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/ed893b6a4e8870d956e3a4c7caae35494fc1aedc)

[6aeb0d1](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/6aeb0d125668a8b3b7f18c21c65a3248b09f6533)

[c9af49f](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/c9af49f5bc92d7927d000e1f8a0a7015d1a0e61d)

test/unit/SavingCirclesUnit.t.sol
```diff
@@ -64,19 +74,111 @@ contract SavingCirclesUnit is Test {
...
+    for (uint256 i = 0; i < _members.length; i++) {
+      address member = _members[i];
+      if (member == _owner) continue;
+
+      bytes memory signature = _signInvite(_circleId, nonce, _ownerKey);
+      vm.prank(member);
+      savingCircles.redeemInvite(_circleId, nonce, signature);
+      nonce++;
+    }
+  }
```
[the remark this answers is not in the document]
exo404: [5201e5f](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/5201e5fbf12872dd19f8a2d163058a14d88b8569)

[bf432fe](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/bf432fe579f5f5734fcebbd813e1cc604b1bf2ea)

[ed893b6](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/ed893b6a4e8870d956e3a4c7caae35494fc1aedc)

[6aeb0d1](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/6aeb0d125668a8b3b7f18c21c65a3248b09f6533)

[c9af49f](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/c9af49f5bc92d7927d000e1f8a0a7015d1a0e61d)

src/contracts/SavingCircles.sol
exo404: The only other option I can think of is adding a size field to the saving circle struct but that would be used only for this and maybe will not reflect the effective number of members so I guess a comment is better

exo404: I think the frontend should generate all the invites at once, after the owner gives the max size input. We can make a validation there checking if the size is less than a constant MAX (maybe 20/25 like the optimal size for safety nets)

exo404: [57afc54](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/57afc54096c6e1bdb5cbfd6f7d81d34649bdfca8)

exo404:
@franrolotti I implemented the changes for the issue #68 in these two commits: [0463444](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/04634440aaa34910bf08eadba01f3f1dcd8fb1fb) [e283738](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/e283738677d4313a2cf91b1698d7330f942928fe)

RonTuretzky:
Check warning on line R148
Check warning:
Possible reentrancy vulnerabilities. Avoid state changes after transfer
@exo404

src/contracts/SavingCircles.sol
```diff
@@ -63,80 +67,113 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    if (_circle.depositInterval > maxDelta / len) revert InvalidDepositInterval();
+    _circle.circleEnd = _circle.effectiveCircleStartTime + (_circle.depositInterval * len);
+
+    isActive[_id] = true;
+    emit CircleStarted(_id);
+  }
+
   /// @inheritdoc ISavingCircles
   function deposit(uint256 _id, uint256 _value) external override nonReentrant {
+    if (!isActive[_id]) revert NotActive();
```
bagelface: could probably make this into an "onlyActive(uint256 _id)" modifier

src/contracts/SavingCircles.sol
```diff
@@ -179,53 +216,67 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
     return _isDecommissioned(_circle);
   }
 
+  function isDecommissionable(uint256 _id) external view override returns (bool) {
+    return _isDecommissionable(_id);
+  }
+
   /// @inheritdoc ISavingCircles
   function isWithdrawable(uint256 _id) public view override returns (bool) {
+    if (!isActive[_id]) revert NotActive();
```
bagelface: this should just return false if it's not active, rather than reverting.

src/contracts/SavingCircles.sol
```diff
@@ -179,53 +216,67 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
     return _withdrawable(_id);
   }
 
   /// @inheritdoc ISavingCircles
   function withdrawableBy(uint256 _id) public view override onlyCommissioned(_id) returns (address) {
+    if (!isActive[_id]) revert NotActive();
```
bagelface: this should just return `address(0)` if it's not active, rather than reverting.

bagelface (commented):
A few small issues and I think we should address the Github generated warnings.

One thing I think we should should consider is that since we don't specify addresses in the signature, the invites can be front run (monitor mempool and send a malicious `redeemtInvite()` transaction that gets cleared before the honest transaction). I don't think this is something that there's any financial incentive to do, so it's not a huge deal but there is potential for abuse here.

src/contracts/SavingCircles.sol
```diff
@@ -295,4 +348,50 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
...
+    return decommissionable;
+  }
+
+  /**
+   * @dev Computes the EIP-712 hash for an invite
+   * @notice _inviteTypehash is keccak256('Invite(uint256 id,uint256 nonce)')
+   * @notice _eip712DomainTypehash is keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)')
+   * @notice Domain name is 'StacksInvite' and version is '1'
+   */
+  function _hashInvite(uint256 _id, uint256 _nonce) private view returns (bytes32) {
```
bagelface, line 387: I don't love the use of "magic numbers" here. Typically, this is done with compile time constants. It's more readable and probably slightly more efficient. I also think we should be implementing OpenZeppelin's `EIP712Upgradeable`.

exo404:
I'll try to address Github linter warnings and warnings in the code (reentracncy pinted by @RonTuretzky) plus the few changes requested by @bagelface in a few hours

src/contracts/SavingCircles.sol
exo404: [18cddc8](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/18cddc863a42c878fd8924d9615c2683dd45a27e)

src/contracts/SavingCircles.sol
exo404: [18cddc8](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/18cddc863a42c878fd8924d9615c2683dd45a27e)

src/contracts/SavingCircles.sol
exo404: [18cddc8](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/18cddc863a42c878fd8924d9615c2683dd45a27e)

exo404:
> Check warning on line R148 Check warning: Possible reentrancy vulnerabilities. Avoid state changes after transfer @exo404

[faead74](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/faead74e07b9fe33737b006f151c1c2f3ac9c8e9)

src/contracts/SavingCircles.sol
exo404: about the magic numbers, first version of the function used constants in the contract but @RonTuretzky said that he prefers to put only the hashes. I personally hate magic numbers too

exo404: I'll implement that interface to address this issue

exo404: [3312a1e](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/3312a1e9450f8c1ad97df79d08a4cd438a78c4c6)

exo404:
this is for github warnings @bagelface [c7388c4](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/c7388c4b2fd80803d7cfbe5f1abf450d3df4942e)

exo404:
[3312a1e](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/3312a1e9450f8c1ad97df79d08a4cd438a78c4c6) @franrolotti

exo404:
@bagelface last warnings here [fix: linter warnings](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/79f01acfec2d971c4a8b212723a21ffe22b31589)

src/contracts/SavingCircles.sol
```diff
@@ -54,6 +58,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
 
   /// @inheritdoc ISavingCircles
   function initialize(address _owner) external override initializer {
+    __EIP712_init(EIP712_NAME, EIP712_VERSION);
     __Ownable_init_unchained(_owner);
```
bagelface: Sorry, just notice this. Why are we using the "unchained" version here? Should probably just call `__Ownable_init(_owner)`.

bagelface (approved):
one last small comment then it looks good to me

src/contracts/SavingCircles.sol
exo404: [714e277](https://github.com/BreadchainCoop/saving-circles/pull/52/commits/714e2779c7c1b322efeb52a077d87f522e2f5f98)

Issue #53, opened 2025-10-22
Assess possibility of dynamic membership and joining mechanism


franrolotti:
Evaluate whether circle members should be able to join dynamically or must be fixed at creation. Define the joining and membership rules before and after `circleStart`, and how these affect deposits, goals, and decommission timing.

* Params: `depositAmount`, `token`, `depositInterval`, `circleStart`, `maxDeposits`.
* Goal = `maxDeposits * depositAmount` * `memberCount`.
* Circle duration = `maxDeposits * depositInterval` after `circleStart`.
* Clarify when parameters and membership become immutable.

Assess membership onboarding with **invite-link–based joining** before the circle start.

RonTuretzky:
Goal = maxdeposits * deposit amount * member count

franrolotti:
@RonTuretzky I understand that we decided to have a fixed member list at the beginning (no dynamic membership after start).
What we need to confirm now is how to achieve that with the invite links.

I’m thinking of the flow as:
**Create circle → generate invites bound to that `circleId` → recipients redeem before `circleStart`.**

To support this, we’d need to allow creating a circle with zero members and enforce the `MINIMUM_MEMBERS` requirement only at or after `circleStart`.
This would let us create a circle even if not all invites are redeemed yet.


What do you think?
@exo404 any input?

RonTuretzky:
> This would let us create a circle even if not all invites are redeemed yet.

sgtm, if having a dynamic membership changes before the circle start is too complex then a constant circle size is also acceptable

PR #58, opened 2025-11-03
feat: hard-deprecate maxDeposits in favor of members.length and align contracts/tests

### LINKED ISSUE #54: Assess whether to deprecate `maxDeposits` in favor of `members.length`
Evaluate removing `maxDeposits` and deriving the number of rounds from `members.length` to guarantee one withdrawal per member.

franrolotti:
Closes #54 

@RonTuretzky I open this draft PR to check if this is really what we needed to do, or if we wanted to do a backward compatible deprecation

### DIFF
--- a/src/contracts/DelegatedSavingCircles.sol
+++ b/src/contracts/DelegatedSavingCircles.sol
@@ -75,7 +75,7 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
         uint256 depositWindowEnd = _circle.circleStart + (_circle.depositInterval * (_circle.currentIndex + 1));
         if (block.timestamp >= depositWindowEnd) continue;
 
-        uint256 circleEnd = _circle.circleStart + (_circle.depositInterval * _circle.maxDeposits);
+        uint256 circleEnd = _circle.circleStart + (_circle.depositInterval * _circle.members.length);
         if (block.timestamp >= circleEnd) continue;
 
         for (uint256 j = 0; j < _circle.members.length; j++) {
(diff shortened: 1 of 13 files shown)


RonTuretzky:
> Closes #54
> 
> @RonTuretzky I open this draft PR to check if this is really what we needed to do, or if we wanted to do a backward compatible deprecation

hard dep no backward compat

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -75,7 +75,7 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
         uint256 depositWindowEnd = _circle.circleStart + (_circle.depositInterval * (_circle.currentIndex + 1));
         if (block.timestamp >= depositWindowEnd) continue;
 
-        uint256 circleEnd = _circle.circleStart + (_circle.depositInterval * _circle.maxDeposits);
+        uint256 circleEnd = _circle.circleStart + (_circle.depositInterval * _circle.members.length);
```
RonTuretzky: i wonder if we should just store this in the circle struct

RonTuretzky (commented):
but you need to also deprecate passing an array into the circle constructor because we're moving to an invite link pattern. I would branch off of that pr for this change

src/contracts/DelegatedSavingCircles.sol
franrolotti: [1690900](https://github.com/BreadchainCoop/saving-circles/pull/58/commits/1690900b7a76431a9dd0f73d4be8130465d70c52)

franrolotti:
@RonTuretzky what do you think if we merge this PR with the `maxDeposits` deprecation and storing `circleEnd` in the struct, and I work on the deprecation of passing an array into the circle constructor because we're moving to an invite link pattern here https://github.com/BreadchainCoop/saving-circles/issues/53

RonTuretzky:
> @RonTuretzky what do you think if we merge this PR with the `maxDeposits` deprecation and storing `circleEnd` in the struct, and I work on the deprecation of passing an array into the circle constructor because we're moving to an invite link pattern here #53

i'm against it , i would rather have all changes self contained in a single pr that also includes the invite link feature , very hard to keep track of everything otherwise

franrolotti:
@RonTuretzky great, just to confirm: I’ll consolidate everything into a single PR branching off of `51-invite-function-integration-from-safety-net`.

* Deprecate passing `members[]` into `create()`.
* Include the other deprecation changes from this PR.

I’ll push these commits to a new branch based on that PR and open a new PR that closes both #51 and the deprecation PR.

PR #62, opened 2025-11-17
merge #54 deprecate maxdeposits in #51 invite function

exo404:


exo404:
@RonTuretzky I'm merging francisco's changes in the invite function PR

PR #70, opened 2025-12-03
fix: optimize start() to use single loop for member filtering

RonTuretzky:
## Summary
- Refactored the `start()` function to eliminate the double loop when filtering null members
- Now uses a single pass to collect non-null members into a temp array, then copies to the correctly sized final array
- Reduces gas costs by only iterating through the full members array once

## Explanation

**Before:** The code had two separate loops over `_circle.members`:
1. First loop counted non-null members to determine the array size
2. Second loop copied non-null members to the new array

**After:** Single loop approach:
1. Allocate a temp array sized to the max possible members
2. In one pass, collect non-null members and count them
3. Copy only `membersCount` elements to the final correctly-sized array

The second copy loop now iterates only `membersCount` times (the actual valid members) instead of the full array length, saving gas when there are null entries.

## Test plan
- [x] Verify existing tests pass
- [ ] Check gas usage improvement in start() function

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -87,23 +87,22 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable {
     Circle storage _circle = circles[_id];
     if (isActive[_id]) revert AlreadyActive();
     if (msg.sender != _circle.owner) revert NotOwner();
+    // Single pass: count and collect non-null members
+    address[] memory tempMembers = new address[](_circle.members.length);
     uint256 membersCount = 0;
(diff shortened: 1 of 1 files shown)


exo404:
This is BEFORE:
image

This is AFTER:
image

exo404:
I see that the only improvement is in the maximum and avg, not that much btw @RonTuretzky. It's a double loop here too so maybe that's the reason

RonTuretzky:
> I see that the only improvement is in the maximum and avg, not that much btw @RonTuretzky. It's a double loop here too so maybe that's the reason


It's less lines and more efficient I think

exo404:
ok I'll just merge this

PR #88, opened 2026-01-12
Fix upgradable issues

### LINKED ISSUE #87: change ReentrancyGuard to his upgradable version
Contract uses non-upgradeable ReentrancyGuard with upgradeable contracts, causing storage slot collision.

exo404:


src/contracts/SavingCircles.sol
```diff
@@ -33,6 +34,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  uint256[30] private __gap;
```
bagelface: why are we adding this gap?

exo404: I wanted to discuss this with you, is this necessary for upgrades to the contract in order to avoid conflict in storage after adding new state variables for example?

bagelface (approved):
lgtm when import conflict is resolved

PR #90, opened 2026-01-12
Use SafeERC20 instead of unsafe functions

### LINKED ISSUE #89: change to SafeERC20
We are using unsafe approve() without SafeERC20, vulnerable to race conditions.

exo404:


bagelface (approved):
lgtm

PR #92, opened 2026-01-12
Storage cleanup

### LINKED ISSUE #91: Storage cleanup
Decommission doesn't clean up isMember, memberCircles, circleMembers mappings.

exo404:


bagelface:
Can you add a description of what the issue was an how this solves it? Is it just that we weren't properly "resetting" values when a circle is decommissioned? If I understand correctly, there aren't really any downsides to leaving this data set, so this just increases gas costs to decommission function without any benefit.

exo404:
> Can you add a description of what the issue was an how this solves it? Is it just that we weren't properly "resetting" values when a circle is decommissioned? If I understand correctly, there aren't really any downsides to leaving this data set, so this just increases gas costs to decommission function without any benefit.

I think it impacts the SavingCircleViewer contract exposing data that aren't valid anymore after a circle is decommissioned

bagelface:
> > Can you add a description of what the issue was an how this solves it? Is it just that we weren't properly "resetting" values when a circle is decommissioned? If I understand correctly, there aren't really any downsides to leaving this data set, so this just increases gas costs to decommission function without any benefit.
> 
> I think it impacts the SavingCircleViewer contract exposing data that aren't valid anymore after a circle is decommissioned

Ya, I guess this can go either way. I personally don't think it's a problem for there to be stale data if the circle is marked as decommissioned. If the viewer is returning incorrect data for some of the view functions it may be better to address the issue there. For example, we could change `_getAllUserCircleIds` to only return circle IDs that aren't decommissioned.

PR #102, opened 2026-01-26
fix: add currentWithdrawer address to the UserCircleData struct

### LINKED ISSUE #101: change the getUserCircleData to return the current withdrawer's address
getUserCircleData returns isCurrentWithdrawer as boolean. Frontend needs the address of the current withdrawer.

exo404:


src/interfaces/ISavingCirclesViewer.sol
```diff
@@ -12,6 +12,7 @@ interface ISavingCirclesViewer {
    * @param isMember Whether the user is a member of the circle
    * @param isOwner Whether the user owns the circle
    * @param isCurrentWithdrawer Whether the user is the current withdrawer
+   * @param currentWithdrawer The address of the current withdrawer
```
franrolotti: `currentWithdrawer` was inserted mid-struct, which change tuple ordering. OK if we are doing a fresh proxy+contract deploy (no upgrade path). If we want to show data from our previous tests we need to append this field at the end.

PR #103, opened 2026-01-26
fix: gate withdrawals to final-day window and update tests

### LINKED ISSUE #98: Move claim availability to the last day of deposit period (maximize token holding time)
**Context**
Today, users can claim/withdraw as soon as the round deposits have been made. We want to **delay claimability to the end of the round’s time window** (the last day) so funds stay in the contract longer.

**Goal**
Make `claim/withdraw` only succeed **when the round’s deposit window fully ends**, instead of being available immediately once deposits have been made.

**Acceptance criteria**

* ✅ `claim()` reverts if called before the last day for that round.
* ✅ `claim()` succeeds on the last day of the round if all required deposits were made.
* ✅ Existing deposit window logic remains unchanged (no late deposits past cutoff).

**Notes / risks to assess**

* UX impact: recipients must wait longer even if all deposits arrive early.
* What happens if they don't claim in that day? They should be able to claim later and delay the start of the next round. Claim for other member is a possibility that minimizes this risk.

franrolotti:
This PR updates withdraw timing so claims are only allowed during the final 24 hours of each round’s deposit window (or later). It also updates unit and integration tests to warp into that last‑day window and to verify late claims still succeed. Note that the round does not advance until a withdrawal happens (`currentIndex` only increments on withdraw) so if a member claims late, the next round is delayed until that withdrawal is made.

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -326,13 +326,16 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
 
   /**
    * @dev Return if a specified circle is withdrawable
-   *      To be considered withdrawable, enough time must have passed since the deposit interval started
-   *      and all members must have made a deposit.
+   *      To be considered withdrawable, the current deposit window must be within its final day
+   *      (last 24 hours of the current round) and all members must have made a deposit.
    */
   function _withdrawable(uint256 _id) internal view onlyCommissioned(_id) returns (bool) {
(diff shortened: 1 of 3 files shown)


RonTuretzky (commented):
.

bagelface:
Why do we want this feature? Honestly, I feel like it might even make sense to just let withdrawals remain available indefinitely? Let users use the saving circle as a sort of bank. If they want use their account for a round deposit, what is the downside of that? Aside from the implication that they didn't really "need" the funds.

PR #107, opened 2026-01-29
Assess time based rounds and allow late claims

### LINKED ISSUE #104: Assess time based rounds: derive `currentIndex` from time; define behavior for unclaimed rounds
**Context / Problem**
Right now we calculate time in two different ways:

* **Round progression** is based on `currentIndex`, which only increments when a **withdrawal/claim** happens. This means claiming early advances the round even if we’re still early in the time window.
* **Deposit windows** are based on fixed timestamps derived from `effectiveCircleStartTime` and `depositInterval`. Time keeps running on schedule even if a withdrawal hasn’t happened yet.

Because those two aren’t aligned, we can end up in weird states:

* **Claim early:** round index moves forward, but the next round’s deposit window ends far in the future.
* **Claim late:** round index is behind, but the next round’s deposit window may have already started or even ended.

We agreed on **A) time-based rounds**: round index is derived from time only; claims do not advance the round (see https://github.com/BreadchainCoop/saving-circles/issues/95).

**Additional context / implication (no-claims execution)**

* **Current behavior:** round progression depends on withdrawals, so the circle **cannot fully execute without anyone claiming** (since `currentIndex` never advances).
* **With time-based rounds:** the round index would advance with time even if nobody claims, so the circle **could “complete” in terms of schedule**.
* **But we must verify:** whether anything else prevents **late claims**, and whether claims are allowed **after the full circle time has passed**.

---

## Goal

1) Implement a time-based round model where the active round index is derived from time, keeping deposit windows and round progression aligned and predictable.
2) Allow late claim any time and end the circle when every claim was made and the contract has no funds left.

franrolotti:
This change makes round progression time-based instead of claim-based. The current round is now derived from `block.timestamp`, and claims no longer advance the round. Deposits are tracked per round, deposit windows align to time, and missed rounds can turn the circle decommissionable. The circle ends only after all members have claimed and the contract has no remaining balance.

Key behavior changes

* Round index is derived from time (`effectiveCircleStartTime + depositInterval`) rather than incrementing on claim.
* Deposits are recorded per round (`roundDeposits`) and member balances for UI/queries reflect the current round only.
* Claims no longer advance `currentIndex`; a claim marks the recipient as claimed and can close the circle once all claims are done and funds are distributed.
* If a prior round is incomplete after its window ends, circle becomes decommissionable and deposits for later rounds revert with `CircleStuck`.
* Decommission refunds only unclaimed rounds.

Follow-up / known limitation

`withdrawableBy` and `isWithdrawable` still only expose the current time-based round. Late claimants can still claim (after their round), but the API doesn’t surface them. We may want to change `withdrawableBy` to return a list of claimable members or add a new function to enumerate claimable recipients. Maybe this could be solved in the frontend.

New storage mappings were added: `hasClaimed`, `lastDepositRound`, and `roundDeposits`. If there’s a way to achieve the same behavior without these additions, let me know and I can adjust.

### DIFF
--- a/src/contracts/DelegatedSavingCircles.sol
+++ b/src/contracts/DelegatedSavingCircles.sol
@@ -150,11 +150,11 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
     // Check if delegated deposits are enabled for this member
     if (!delegatedDepositsEnabled[_member]) revert DelegatedDepositsNotEnabled();
 
+    if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.NotActive();
+
     // Get circle information
     ISavingCircles.Circle memory _circle = SAVING_CIRCLES.getCircle(_circleId);
     address[] memory circleMembers = SAVING_CIRCLES.getCircleMembers(_circleId);
-
-    if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.NotActive();
 
     // Check if member is part of the circle
     bool isMember = false;
--- a/src/contracts/DelegatedSavingCircles.sol
+++ b/src/contracts/DelegatedSavingCircles.sol
@@ -77,17 +77,17 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
         // Check if we're in a valid deposit window
         if (block.timestamp < _circle.effectiveCircleStartTime) continue;
 
-        uint256 depositWindowEnd =
-          _circle.effectiveCircleStartTime + (_circle.depositInterval * (_circle.currentIndex + 1));
-        if (block.timestamp >= depositWindowEnd) continue;
+        if (SAVING_CIRCLES.isDecommissionable(id)) continue;
 
-        if (block.timestamp >= _circle.circleEnd) continue;
+        uint256 currentRound = _currentRoundIndex(_circle);
+        if (currentRound >= circleMembers.length) continue;
 
+        (, uint256[] memory memberBalances) = SAVING_CIRCLES.getMemberBalances(id);
         for (uint256 j = 0; j < circleMembers.length; j++) {
           address member = circleMembers[j];
           if (!delegatedDepositsEnabled[member]) continue; // Skip if not opted in
 
-          uint256 currentBalance = SAVING_CIRCLES.balances(id, member);
+          uint256 currentBalance = memberBalances[j];
           if (currentBalance >= _circle.depositAmount) continue; // Already deposited
 
           uint256 requiredAmount = _circle.depositAmount - currentBalance;
@@ -114,17 +114,15 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
         // Check if we're in a valid deposit window
         if (block.timestamp < _circle.effectiveCircleStartTime) continue;
 
-        uint256 depositWindowEnd =
-          _circle.effectiveCircleStartTime + (_circle.depositInterval * (_circle.currentIndex + 1));
-        if (block.timestamp >= depositWindowEnd) continue;
+        uint256 currentRound = _currentRoundIndex(_circle);
+        if (currentRound >= circleMembers.length) continue;
 
-        if (block.timestamp >= _circle.circleEnd) continue;
-
+        (, uint256[] memory memberBalances) = SAVING_CIRCLES.getMemberBalances(id);
         for (uint256 j = 0; j < circleMembers.length; j++) {
           address member = circleMembers[j];
           if (!delegatedDepositsEnabled[member]) continue; // Skip if not opted in
 
-          uint256 currentBalance = SAVING_CIRCLES.balances(id, member);
+          uint256 currentBalance = memberBalances[j];
           if (currentBalance >= _circle.depositAmount) continue; // Already deposited
 
           uint256 requiredAmount = _circle.depositAmount - currentBalance;
@@ -156,6 +154,8 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
     ISavingCircles.Circle memory _circle = SAVING_CIRCLES.getCircle(_circleId);
     address[] memory circleMembers = SAVING_CIRCLES.getCircleMembers(_circleId);
 
+    if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.NotActive();
+
     // Check if member is part of the circle
     bool isMember = false;
     for (uint256 i = 0; i < circleMembers.length; i++) {
@@ -167,7 +167,7 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
     if (!isMember) revert ISavingCircles.NotMember();
 
     // Calculate the remaining deposit amount needed
-    uint256 currentBalance = SAVING_CIRCLES.balances(_circleId, _member);
+    uint256 currentBalance = _memberBalance(_circleId, _member);
     if (currentBalance >= _circle.depositAmount) revert ISavingCircles.AlreadyDeposited();
     uint256 requiredAmount = _circle.depositAmount - currentBalance;
 
@@ -179,12 +179,8 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
     if (block.timestamp < _circle.effectiveCircleStartTime) {
       revert ISavingCircles.DepositBeforeCircleStart();
     }
-    // Check if current deposit window has closed (each window lasts depositInterval)
-    if (block.timestamp >= _circle.effectiveCircleStartTime + (_circle.depositInterval * (_circle.currentIndex + 1))) {
-      revert ISavingCircles.DepositWindowClosed();
-    }
     // Check if all deposit periods have passed
-    if (block.timestamp >= _circle.circleEnd) {
+    if (_currentRoundIndex(_circle) >= circleMembers.length) {
       revert ISavingCircles.CircleExpired();
     }
 
@@ -196,5 +192,26 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
 
     // Call depositFor on the main contract
     SAVING_CIRCLES.depositFor(_circleId, requiredAmount, _member);
+  }
+
+  function _currentRoundIndex(ISavingCircles.Circle memory _circle) internal view returns (uint256) {
+    if (
+      _circle.depositInterval == 0 || _circle.effectiveCircleStartTime == 0
+        || block.timestamp < _circle.effectiveCircleStartTime
+    ) {
+      return 0;
+    }
+
+    return (block.timestamp - _circle.effectiveCircleStartTime) / _circle.depositInterval;
+  }
+
+  function _memberBalance(uint256 _circleId, address _member) internal view returns (uint256) {
+    (address[] memory members, uint256[] memory balances) = SAVING_CIRCLES.getMemberBalances(_circleId);
+    for (uint256 i = 0; i < members.length; i++) {
+      if (members[i] == _member) {
+        return balances[i];
+      }
+    }
+    return 0;
   }
 }
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -36,6 +36,9 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => mapping(address member => bool claimed)) public hasClaimed;
+  mapping(uint256 id => mapping(address member => uint256 round)) private lastDepositRound;
+  mapping(uint256 id => mapping(uint256 round => mapping(address member => uint256 amount))) public roundDeposits;
 
   /// @dev Requires circle is commissioned by checking if an owner is set
   modifier onlyCommissioned(uint256 _id) {
@@ -186,6 +189,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   /// @inheritdoc ISavingCircles
   function getCircle(uint256 _id) external view override onlyCommissioned(_id) returns (Circle memory _circle) {
     _circle = circles[_id];
+    _circle.currentIndex = _currentRoundIndex(_circle);
   }
 
   /// @inheritdoc ISavingCircles
@@ -194,6 +198,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
 
     for (uint256 i = 0; i < _ids.length; i++) {
       _circles[i] = circles[_ids[i]];
+      _circles[i].currentIndex = _currentRoundIndex(_circles[i]);
     }
   }
 
@@ -229,9 +234,15 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
 
     if (_isDecommissioned(_circle)) revert NotCommissioned();
 
+    uint256 currentRound = _currentRoundIndex(_circle);
     _balances = new uint256[](circleMembers[_id].length);
     for (uint256 i = 0; i < circleMembers[_id].length; i++) {
-      _balances[i] = balances[_id][circleMembers[_id][i]];
+      address member = circleMembers[_id][i];
+      if (lastDepositRound[_id][member] == currentRound) {
+        _balances[i] = balances[_id][member];
+      } else {
+        _balances[i] = 0;
+      }
     }
 
     return (circleMembers[_id], _balances);
@@ -254,7 +265,11 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   /// @inheritdoc ISavingCircles
   function isWithdrawable(uint256 _id) public view override returns (bool) {
     if (!isActive[_id]) return false;
-    return _withdrawable(_id);
+    Circle memory _circle = circles[_id];
+    uint256 currentRound = _currentRoundIndex(_circle);
+    if (currentRound >= circleMembers[_id].length) return false;
+    address member = circleMembers[_id][currentRound];
+    return _claimable(_id, member);
   }
 
   /// @inheritdoc ISavingCircles
@@ -262,7 +277,9 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     if (!isActive[_id]) return address(0);
     Circle memory _circle = circles[_id];
 
-    return circleMembers[_id][_circle.currentIndex];
+    uint256 currentRound = _currentRoundIndex(_circle);
+    if (currentRound >= circleMembers[_id].length) return address(0);
+    return circleMembers[_id][currentRound];
   }
 
   /**
@@ -273,20 +290,19 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   function _withdraw(uint256 _id, address _member) internal onlyMember(_id, msg.sender) {
     Circle storage _circle = circles[_id];
 
-    if (!_withdrawable(_id)) revert NotWithdrawable();
-    if (circleMembers[_id][_circle.currentIndex] != _member) revert NotWithdrawable();
-    if (_circle.currentIndex >= circleMembers[_id].length) revert NotWithdrawable();
+    if (!_claimable(_id, _member)) revert NotWithdrawable();
 
     uint256 _withdrawAmount = _circle.depositAmount * (circleMembers[_id].length);
 
-    for (uint256 i = 0; i < circleMembers[_id].length; i++) {
-      balances[_id][circleMembers[_id][i]] = 0;
-    }
+    hasClaimed[_id][_member] = true;
 
-    _circle.currentIndex = (_circle.currentIndex + 1) % circleMembers[_id].length;
     IERC20(_circle.token).safeTransfer(_member, _withdrawAmount);
 
     emit FundsWithdrawn(_id, _member, _withdrawAmount);
+
+    if (_allMembersClaimed(_id) && IERC20(_circle.token).balanceOf(address(this)) == 0) {
+      isActive[_id] = false;
+    }
   }
 
   /**
@@ -304,22 +320,29 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     if (block.timestamp < circles[_id].effectiveCircleStartTime) {
       revert DepositBeforeCircleStart();
     }
-    // Check if the entire circle has expired (all rounds completed)
-    if (block.timestamp >= _circle.circleEnd) {
+    uint256 currentRound = _currentRoundIndex(_circle);
+    if (currentRound >= circleMembers[_id].length) {
       revert CircleExpired();
     }
-    // Check if current deposit window is closed
-    if (
-      block.timestamp
-        >= circles[_id].effectiveCircleStartTime + (circles[_id].depositInterval * (circles[_id].currentIndex + 1))
-    ) {
-      revert DepositWindowClosed();
-    }
-    if (balances[_id][_member] + _value > circles[_id].depositAmount) {
-      revert ExceedsDepositAmount();
+
+    if (currentRound > 0) {
+      uint256 prev = currentRound - 1;
+      if (
+        block.timestamp > _roundEndTime(_circle, prev)
+          && !_allMembersDepositedForRound(_id, prev, _circle.depositAmount)
+      ) {
+        revert CircleStuck();
+      }
     }
 
-    balances[_id][_member] = balances[_id][_member] + _value;
+    uint256 depositedSoFar = roundDeposits[_id][currentRound][_member];
+    if (depositedSoFar + _value > _circle.depositAmount) revert ExceedsDepositAmount();
+
+    uint256 newTotal = depositedSoFar + _value;
+    roundDeposits[_id][currentRound][_member] = newTotal;
+
+    lastDepositRound[_id][_member] = currentRound;
+    balances[_id][_member] = newTotal;
 
     IERC20(_circle.token).safeTransferFrom(msg.sender, address(this), _value);
 
@@ -327,21 +350,19 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   }
 
   /**
-   * @dev Return if a specified circle is withdrawable
-   *      To be considered withdrawable, enough time must have passed since the deposit interval started
-   *      and all members must have made a deposit.
+   * @dev Return if a specified member can claim from a circle
+   *      Claim eligibility is determined by the time-based round index.
    */
-  function _withdrawable(uint256 _id) internal view onlyCommissioned(_id) returns (bool) {
+  function _claimable(uint256 _id, address _member) internal view onlyCommissioned(_id) returns (bool) {
     Circle memory _circle = circles[_id];
+    if (hasClaimed[_id][_member]) return false;
 
-    if (block.timestamp < _circle.effectiveCircleStartTime + (_circle.depositInterval * _circle.currentIndex)) {
-      return false;
-    }
-    address[] memory members = circleMembers[_id];
-    for (uint256 i = 0; i < members.length; i++) {
-      if (balances[_id][members[i]] < _circle.depositAmount) {
-        return false;
-      }
+    uint256 currentRound = _currentRoundIndex(_circle);
+    (uint256 memberIndex, bool found) = _memberIndex(_id, _member);
+    if (!found || currentRound < memberIndex) return false;
+
+    if (currentRound == memberIndex) {
+      return _allMembersDepositedForRound(_id, currentRound, _circle.depositAmount);
     }
 
     return true;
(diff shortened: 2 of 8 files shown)


src/contracts/DelegatedSavingCircles.sol
```diff
@@ -197,4 +193,25 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
...
+
+  function _currentRoundIndex(ISavingCircles.Circle memory _circle) internal view returns (uint256) {
+    if (
+      _circle.depositInterval == 0 || _circle.effectiveCircleStartTime == 0
+        || block.timestamp < _circle.effectiveCircleStartTime
+    ) {
+      return 0;
+    }
+
+    return (block.timestamp - _circle.effectiveCircleStartTime) / _circle.depositInterval;
```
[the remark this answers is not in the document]
RonTuretzky: Lets address this in a seperate PR

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -197,4 +193,25 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
...
+  }
+
+  function _memberBalance(uint256 _circleId, address _member) internal view returns (uint256) {
+    (address[] memory members, uint256[] memory balances) = SAVING_CIRCLES.getMemberBalances(_circleId);
+    for (uint256 i = 0; i < members.length; i++) {
+      if (members[i] == _member) {
+        return balances[i];
+      }
+    }
+    return 0;
```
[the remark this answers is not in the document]
RonTuretzky: Lets address this in a seperate PR

src/contracts/SavingCircles.sol
```diff
@@ -138,18 +141,26 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
     address[] memory members = circleMembers[_id];
+    uint256 len = members.length;
+
     isActive[_id] = false;
 
-    // Return deposits to members
-    for (uint256 i = 0; i < members.length; i++) {
-      address _member = members[i];
-      uint256 _balance = balances[_id][_member];
+    // Refund all deposits in rounds whose payout hasn't happened yet (recipient not claimed)
```
RonTuretzky: what about if they deposited but its not their turn to claim? why not just refund everyone their balance as it was before in addition to doing their claims?

src/contracts/SavingCircles.sol
```diff
@@ -378,6 +399,56 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
     return _circle.owner == address(0);
   }
 
+  function _roundEndTime(Circle memory _circle, uint256 round) internal pure returns (uint256) {
+    return _circle.effectiveCircleStartTime + (_circle.depositInterval * (round + 1));
+  }
+
+  function _currentRoundIndex(Circle memory _circle) internal view returns (uint256) {
```
RonTuretzky, line 416: this function doesn't make sense to me, doesn't this always return 0?

src/contracts/SavingCircles.sol
```diff
@@ -378,6 +399,56 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
+    for (uint256 i = 0; i < members.length; i++) {
+      address member = members[i];
+      if (roundDeposits[_id][_round][member] < _depositAmount) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  function _memberIndex(uint256 _id, address _member) internal view returns (uint256, bool) {
```
RonTuretzky: maybe we should add a mapping for this

RonTuretzky (changes requested):
Please address warning as well

src/contracts/SavingCircles.sol
franrolotti: @RonTuretzky i think i don't understand this comment...

What we are doing here is to return what users deposited in previous rounds, and that haven't been withdrawn yet (so the funds are still in the contract).

If your concern is members who deposited into earlier rounds but haven’t reached their payout yet, agreed they can end up net-negative after decommission.

Could you please clarify what you mean here?

src/contracts/SavingCircles.sol
franrolotti: It returns 0 only if the time params aren’t initialized (`depositInterval == 0` or `effectiveCircleStartTime == 0`) or if we’re before `effectiveCircleStartTime`.
Once `block.timestamp >= effectiveCircleStartTime`, it returns the number of full `depositIntervals` elapsed since start `((now - start)/interval)`

src/contracts/SavingCircles.sol
```diff
@@ -271,20 +296,19 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
-      balances[_id][circleMembers[_id][i]] = 0;
-    }
+    hasClaimed[_id][_member] = true;
 
-    _circle.currentIndex = (_circle.currentIndex + 1) % circleMembers[_id].length;
     IERC20(_circle.token).safeTransfer(_member, _withdrawAmount);
 
     emit FundsWithdrawn(_id, _member, _withdrawAmount);
+
+    if (_allMembersClaimed(_id) && IERC20(_circle.token).balanceOf(address(this)) == 0) {
```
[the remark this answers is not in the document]
franrolotti: [4eadace](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/4eadaceec872829e03fa873558b95bb47ee3cf2a)

src/contracts/SavingCircles.sol
```diff
@@ -302,73 +326,70 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
-    ) {
-      revert DepositWindowClosed();
-    }
-    if (balances[_id][_member] + _value > circles[_id].depositAmount) {
-      revert ExceedsDepositAmount();
+
+    if (currentRound > 0) {
+      uint256 prev = currentRound - 1;
+      if (
+        block.timestamp > _roundEndTime(_circle, prev)
```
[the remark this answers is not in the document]
franrolotti: [48cd38a](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/48cd38aff4fc50e028264a46ca1ba27eb70ed010)

src/contracts/SavingCircles.sol
```diff
@@ -302,73 +326,70 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
-      }
-    }
-    if (!hasIncompleteDeposits) decommissionable = false;
+    if (currentRound == 0) return false;
+
+    uint256 checkRound = currentRound - 1;
 
-    return decommissionable;
+    if (checkRound >= len) checkRound = len - 1;
+    if (block.timestamp <= _roundEndTime(_circle, checkRound)) return false;
```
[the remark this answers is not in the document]
franrolotti: [48cd38a](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/48cd38aff4fc50e028264a46ca1ba27eb70ed010)

src/contracts/SavingCircles.sol
franrolotti: [0cb4fd5](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/0cb4fd5b89c3e9d6e1a77f4b62b11c3cd71ed1d6)

src/contracts/SavingCircles.sol
```diff
@@ -38,6 +38,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
   mapping(uint256 id => mapping(address member => bool claimed)) public hasClaimed;
   mapping(uint256 id => mapping(address member => uint256 round)) private lastDepositRound;
   mapping(uint256 id => mapping(uint256 round => mapping(address member => uint256 amount))) public roundDeposits;
+  mapping(uint256 id => mapping(address member => uint256 indexPlusOne)) private memberIndexPlusOne;
```
RonTuretzky: I would prefer to call this member index and add in a comment that its an index that starts at 1

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
RonTuretzky: if indeed that is what we are doing then this comment is misleading. The comment implies we are only return people who have a valid claim, leaving all funds that are currently in the contract (that are not claimable) to be permanantly stuck in the contract.

src/contracts/SavingCircles.sol
```diff
@@ -378,6 +399,56 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
...
+
+  function _currentRoundIndex(Circle memory _circle) internal view returns (uint256) {
+    if (
+      _circle.depositInterval == 0 || _circle.effectiveCircleStartTime == 0
+        || block.timestamp < _circle.effectiveCircleStartTime
+    ) {
+      return 0;
+    }
+
+    return (block.timestamp - _circle.effectiveCircleStartTime) / _circle.depositInterval;
```
RonTuretzky, line 424: don't we need a precision adjustment here?

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
franrolotti: Here is intentionally used integer division to get a whole number round index (floor), so I think that no additional precision adjustment is needed. Am I missing something?

src/contracts/SavingCircles.sol
franrolotti: [c0887e9](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/c0887e9397751d20ce8ec68f5e94524a0248c18e) wdyt?

src/contracts/SavingCircles.sol
franrolotti: [682b4ce](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/682b4ce09a92fbbdc81b3431a432b870e86e547b)

src/contracts/SavingCircles.sol
RonTuretzky: no sounds like you're right

src/interfaces/ISavingCircles.sol
```diff
@@ -203,6 +203,10 @@ interface ISavingCircles {
    * @notice Thrown when the circle is not active
    */
   error NotActive();
+  /**
+   * @notice Thrown when a deposit is missed
+   */
+  error CircleStuck();
```
bagelface: `CircleTimeout` or `CircleTimedOut` feels like a better description of how this error is used. "Stuck" to me has a different connotation in context of smart contracts.

src/contracts/SavingCircles.sol
```diff
@@ -35,6 +35,11 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => mapping(address member => bool claimed)) public hasClaimed;
+  mapping(uint256 id => mapping(address member => uint256 round)) private _lastDepositRound;
```
bagelface: would move this private variable down to the bottom of the declarations to keep it organized.

src/contracts/SavingCircles.sol
```diff
@@ -140,16 +146,23 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
 
     address token = circles[_id].token;
     address[] memory members = circleMembers[_id];
+    uint256 len = members.length;
```
bagelface: this should be "membersLength" otherwise this variable name is too ambiguous

src/contracts/DelegatedSavingCircles.sol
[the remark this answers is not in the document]
bagelface: why would we wait to fix this in a separate PR? am I overlooking something that would make that complicated? this seems like a 5 minute change tops

src/contracts/DelegatedSavingCircles.sol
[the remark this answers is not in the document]
bagelface: same question here about waiting to address this
https://github.com/BreadchainCoop/saving-circles/pull/107/changes#r2784075413

src/contracts/SavingCircles.sol
RonTuretzky: tbh i dont even like having private variables at all for this

bagelface (changes requested):
I think with the big change of allowing members to claim from past rounds, we probably should add a few more tests to ensure there isn't any issues being overlooked. Some suggested tests:

1. Member with `memberIndex > 0` tries to withdraw during round 0 — should fail via `currentRound < memberIndex`
2. Time skips to round N, all prior-round members claim in one block — should all succeed
3. Time skips to round N, a future-round member tries to claim — should fail
4. Member claims late (their round is long past) — should still succeed if deposits for their round were complete

I'd recommend adding targeted tests for these four scenarios to be confident that the timestamp-based approach doesn't introduce regressions.

src/interfaces/ISavingCircles.sol
franrolotti: [fef1812](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/fef1812895459a1e8a6c92634252c56e1ad86b0c)

src/contracts/SavingCircles.sol
franrolotti: [1fc5a55](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/1fc5a555be6af4d68cabe57a3e9b09495a3a5410)

src/contracts/SavingCircles.sol
franrolotti: [c6e5d97](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/c6e5d97674c652e861fd3b55594d6a569674580b)

franrolotti:
@RonTuretzky @bagelface added the following tests here [5a0a694](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/5a0a694b03a7db3a3634032e3a26dafddb121339)

* `test_WithdrawFutureRoundMemberReverts`: verifies a member cannot withdraw before their assigned round.
* `test_WithdrawPastRoundClaimsSucceedSameBlock`: verifies that after skipping to a later round, eligible past-round members can still claim, even in the same block.
* `test_WithdrawSkipToRoundFutureMemberReverts`: verifies that skipping time forward does not let future-round members claim early.
* `test_WithdrawLateClaimSucceeds`: verifies a member can claim late (well after their round) when their round deposits were complete.
* `test_WithdrawLateClaimRevertsWhenCircleBecomesDecommissionable`: verifies late claims are blocked once the circle enters decommissionable state.
* `test_WithdrawAllMembersOutOfOrderThenCircleDeactivates`: verifies all members can claim out of order once eligible, and the circle deactivates after all claims.

Let me know if you think about other edge cases to cover, and I can add them in this PR.

src/contracts/SavingCircles.sol
```diff
@@ -88,6 +93,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
     isMember[_id][owner] = true;
     memberCircles[owner].push(_id);
     circleMembers[_id].push(owner);
+    memberIndex[_id][owner] = circleMembers[_id].length;
```
bagelface: Ron had a comment on a PR that makes sense to me. This should just be indexed from 0 with the 0 index being the circle owner. Every circle has an owner. If we need some kind of extra validation on querying "memberIndex" like if the circle is decommissioned or something, then it should be an internal mapping with an explicit "getter" function.

exo404: I agree

test/unit/SavingCirclesUnit.t.sol
```diff
@@ -208,6 +208,168 @@ contract SavingCirclesUnit is SavingCirclesTestBase {
     savingCircles.withdraw(baseCircleId);
   }
 
+  function test_WithdrawFutureRoundMemberReverts() external {
```
bagelface, line 211: nice new tests look good

src/contracts/SavingCircles.sol
franrolotti: Makes total sense [79086da](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/79086da71d3530b9fcf831b7bc73a2ae038e7905)

RonTuretzky: @franrolotti am I misunderstanding the code or did you put the owner in the N-1 spot?

franrolotti: @RonTuretzky  We set it as `memberIndex[_id][owner] = circleMembers[_id].length - 1` to keep indexing logic consistent with the rest of the contract. Index is always derived from "position after push," including in `redeemInvite`.
In create, this evaluates to `0` anyway because owner is the first member pushed, so behavior is the same today. This was covered by `test_MemberIndexIsZeroBasedAndOwnerFirst`, which explicitly asserts owner index is `0` and member indexing is zero-based.

If you prefer, I can change it to hardcoded `0` for the owner.

RonTuretzky: yea I do prefer that , this is not readable and has extra arithmetic ops we dont need  🙏

franrolotti: makes sense [a2dd72d](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/a2dd72d8e8599eee568a777160c8169e7605077f)

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -156,6 +154,8 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
     ISavingCircles.Circle memory _circle = SAVING_CIRCLES.getCircle(_circleId);
     address[] memory circleMembers = SAVING_CIRCLES.getCircleMembers(_circleId);
 
+    if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.NotActive();
```
RonTuretzky: shouldn't this be on L152?

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -156,6 +154,8 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
...
+
```
RonTuretzky:
```suggestion
if (!delegatedDepositsEnabled[_member]) revert DelegatedDepositsNotEnabled();
if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.NotActive();

// Get circle information
ISavingCircles.Circle memory _circle = SAVING_CIRCLES.getCircle(_circleId);
address[] memory circleMembers = SAVING_CIRCLES.getCircleMembers(_circleId);


```

src/contracts/SavingCircles.sol
```diff
@@ -36,6 +36,10 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => mapping(address member => bool claimed)) public hasClaimed;
+  mapping(uint256 id => mapping(address member => uint256 round)) public lastDepositRound;
+  mapping(uint256 id => mapping(uint256 round => mapping(address member => uint256 amount))) public roundDeposits;
+  mapping(uint256 id => mapping(address member => uint256 memberIndex)) public memberIndex;
```
RonTuretzky: looks like L39-L40-L42 can all be conslidated to a struct , possible to even conslidate L41 , not sure

src/contracts/SavingCircles.sol
```diff
@@ -36,6 +36,10 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => mapping(address member => bool claimed)) public hasClaimed;
```
RonTuretzky: isn't this per round though?

test/unit/SavingCirclesViewerUnit.t.sol
```diff
@@ -216,9 +216,6 @@ contract SavingCirclesViewerUnit is SavingCirclesTestBase {
     savingCircles.deposit(baseCircleId, DEPOSIT_AMOUNT);
     vm.stopPrank();
 
-    // Move time past first round to enable withdrawal
-    vm.warp(block.timestamp + DEPOSIT_INTERVAL);
-
```
RonTuretzky, line 221: why is this being removed?

RonTuretzky (changes requested):

src/contracts/SavingCircles.sol
franrolotti: `hasClaimed` is deliberately per cirlce (one payout per member per circle). Per-round tracking is handled by `roundDeposits`.

test/unit/SavingCirclesViewerUnit.t.sol
franrolotti: This warp was a leftover from an earlier test version and was unnecessary. Once all members deposit in the current round, the current withdrawer can withdraw without advancing the round.

src/contracts/SavingCircles.sol
```diff
@@ -254,15 +275,21 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
...
 
   /// @inheritdoc ISavingCircles
   function withdrawableBy(uint256 _id) public view override onlyCommissioned(_id) returns (address) {
     if (!isActive[_id]) return address(0);
     Circle memory _circle = circles[_id];
 
-    return circleMembers[_id][_circle.currentIndex];
+    uint256 currentRound = _currentRoundIndex(_circle);
+    if (currentRound >= circleMembers[_id].length) return address(0);
+    return circleMembers[_id][currentRound];
```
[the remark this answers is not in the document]
franrolotti: Expected behavior here is that eligibility is not limited to the exact current round: a member can still claim in later rounds if their payout is claimable and unclaimed. 

Given that, `withdrawableBy(id)`serves a different purpose than `isWithdrawable(id)`: it indicates the scheduled member for the current round (turn/order visibility), while `isWithdrawable(id)` checks whether a withdrawal is actually executable now.

So these two functions intentionally expose different signals:

* `withdrawableBy` = round schedule / “whose turn is it”
* `isWithdrawable` = executable claim state

Applying Copilot’s suggestion would merge those semantics and remove useful state information used by the viewer. Wdyt??

src/contracts/SavingCircles.sol
```diff
@@ -142,16 +147,23 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
...
+        if (amount == 0) continue;
+
+        roundDeposits[_id][r][member] = 0;
 
-      if (_balance > 0) {
-        balances[_id][_member] = 0;
-        IERC20(token).safeTransfer(_member, _balance);
+        IERC20(token).safeTransfer(member, amount);
       }
     }
```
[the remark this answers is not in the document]
franrolotti: @RonTuretzky This is very important. I think the current approach is correct for refunding actual per-round deposits, but I agree there’s optimization room. Since we need to merge this to unblock testing for the next round, is it good if we open to a follow-up PR focused on optimizing this?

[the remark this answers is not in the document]
franrolotti: I double-checked this path and I think the current logic is correct.

In this contract, round index and recipient index are intentionally aligned (round `r` pays `members[r]`).
So if (`_memberStates[_id][members[r]].hasClaimed`) continue,  means that payout round was already settled and should not be refunded again during decommission.

If we refunded those rounds again, we’d risk double-paying funds that were already transferred in `withdraw`.
This behavior is also in `testFuzz_PartialRoundsWithDecommission`, where earlier rounds are claimed and later incomplete rounds are refunded on decommission.

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -197,4 +193,25 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
...
+    return (block.timestamp - _circle.effectiveCircleStartTime) / _circle.depositInterval;
+  }
+
+  function _memberBalance(uint256 _circleId, address _member) internal view returns (uint256) {
+    (address[] memory members, uint256[] memory balances) = SAVING_CIRCLES.getMemberBalances(_circleId);
+    for (uint256 i = 0; i < members.length; i++) {
+      if (members[i] == _member) {
+        return balances[i];
+      }
+    }
```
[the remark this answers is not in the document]
franrolotti: @RonTuretzky previously we talked about leaving this one for a separate PR. Do you want me to do it now or do it afterwards in the issues you created?

src/contracts/DelegatedSavingCircles.sol
```diff
@@ -197,4 +193,25 @@ contract DelegatedSavingCircles is IDelegatedSavingCircles, ReentrancyGuard {
...
+    return 0;
+  }
```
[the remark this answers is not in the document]
franrolotti: same question than in the previous one @RonTuretzky

src/contracts/DelegatedSavingCircles.sol
franrolotti: [2053288](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/2053288fe865ffcd550ed5882b89af472dab8077)

src/contracts/DelegatedSavingCircles.sol
franrolotti: [2053288](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/2053288fe865ffcd550ed5882b89af472dab8077)

src/contracts/SavingCircles.sol
franrolotti: [146b099](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/146b0996a64700aa0e7b5fb0e44ae23a548b9fd5)

src/interfaces/ISavingCircles.sol
```diff
@@ -203,6 +203,10 @@ interface ISavingCircles {
    * @notice Thrown when the circle is not active
    */
   error NotActive();
+  /**
+   * @notice Thrown when a deposit is missed
```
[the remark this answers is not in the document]
franrolotti: [20133ee](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/20133ee21bbd71df45a5c6c5b2cab96c7e1ddcfe)

src/contracts/SavingCircles.sol
```diff
@@ -304,80 +330,122 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
...
-  function _isDecommissioned(Circle memory _circle) internal pure returns (bool) {
-    return _circle.owner == address(0);
+    return _allMembersDepositedForRound(_id, memberIndex, _circle.depositAmount);
   }
 
   /**
    * @dev Return if a specified circle is decommissionable
    *      To be considered decommissionable, the circle must have passed its deposit window
-   *      and all members must have made their deposits for the current round.
+   *      and some members must have incomplete deposits for the current round.
```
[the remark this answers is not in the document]
franrolotti: [abb917c](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/abb917c14a4901e5e9945d136379ec33b3eb6b55)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: sgtm

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: i think you should just rename "withdrawableBy" to "currentClaimer"

src/contracts/DelegatedSavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: no lets leave it for a different pr

src/contracts/DelegatedSavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: leave it

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
bagelface: might be good to change the variable name a bit to more descriptive? `currentRoundWithdrawer` or `currentRoundWithdrawableBy` maybe?

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
bagelface: Do we need to refund anything but the unfinished round? Maybe I'm misunderstanding this, but it seems like this will refund from previous rounds if people haven't explicitly claimed them, which doesn't seem ideal.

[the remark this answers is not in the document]
franrolotti: Current decommission behavior is intentional for now: it returns all funds still held by the contract back to the original depositors, instead of only handling the unfinished round. The rationale is to avoid outcomes where one past-round claimant gets paid while N-1 contributors from that round are left worse off after decommission.

I agree this is a policy choice with tradeoffs and probably needs a proper spec. To avoid blocking this PR, I propose we keep the current behavior and open a follow-up issue to define and optimize decommission logic for next version. 

@bagelface @RonTuretzky

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
franrolotti: [9020b55](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/9020b557d2f66a2d1d37d3870658515dcd32dfc2)

src/contracts/SavingCircles.sol
```diff
@@ -179,13 +195,15 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     isMember[_id][msg.sender] = true;
     memberCircles[msg.sender].push(_id);
     circleMembers[_id].push(msg.sender);
+    _memberStates[_id][msg.sender].memberIndex = circleMembers[_id].length - 1;
```
RonTuretzky: nit: we're dereferncing the `circleMembers[_id]` struct here twice

RonTuretzky (approved):

src/contracts/SavingCircles.sol
franrolotti: [451a695](https://github.com/BreadchainCoop/saving-circles/pull/107/commits/451a695588cad07784eadc76a4e30ffe4e135524)

franrolotti:
| Item | Details |
|------|---------|
| Network | `Gnosis` (`chainId 100`) |
| `SavingCircles` Proxy (frontend) | `0x7bC9212FA4Ad242D5Cee496e0570b312446877E3` |
| `SavingCircles` Implementation | `0x58Ad17Fe5Db2a4228942991BCA1bB7A1ff5e54a1` |
| `ProxyAdmin` | `0x498C6b190130CEb23F4Fab1C3d19388F5D55b67b` |
| `DelegatedSavingCircles` | `0x6DA3dAFa3d1372458eefa5707dfb1ED0cE19ab4d` |
| `SavingCirclesViewer` | `0x30142762922fa1594eA0b9e2e9a3b167F5FF31B0` |
| Allowed token (`BREAD`) | `0xa555d5344f6FB6c65da19e403Cb4c1eC4a1a5Ee3` |

PR #117, opened 2026-02-09
Fix decommissioned circle data access in viewer

### LINKED ISSUE #112: `decommission` deletes circle data, causing `getUserCircleData` to revert
`SavingCircles.decommission` deletes `circles[_id]`, and `getCircle` is guarded by `onlyCommissioned`. This makes `SavingCirclesViewer.getUserCircleData` revert for decommissioned circles, so the frontend can’t show historical circle details. 

Frontend should still fetch circle details after decommission.

franrolotti:
## Summary
- Keep circle storage on decommission while still marking decommissioned by zeroing `owner`.
- Store the original owner in `circleOwners` so historical data can be displayed.
- Update the viewer to reconstruct circles from raw storage so it won’t revert on decommissioned circles.
- Fix fuzz test expectation for the zero/zero creation edge case.

## Issue
The current `decommission` deletes `circles[_id]` and `getCircle` is guarded by `onlyCommissioned`. This makes `SavingCirclesViewer.getUserCircleData` revert for decommissioned circles, so the frontend can’t show historical circle details. This PR keeps the data and restores viewer access.

## Notes / Possible flaws
- If this upgrade is applied to the existing deployment, older circles won’t have `circleOwners` populated; after decommission, their original owner may be lost unless migrated.
- Storage grows over time because circle records are no longer deleted.
- Viewer now bypasses the `onlyCommissioned` guard by reading raw storage, which is intentional for history but changes the access pattern.

## Alternative approach
- Drop the `owner == address(0)` meaning and add an explicit `isDecommissionedById` flag, optionally storing only a minimal snapshot for display. This is cleaner long‑term but requires changing semantics.
- Alternatively, remove `onlyCommissioned` from `getCircle` and keep circle data; then the viewer can call `getCircle` directly without bypassing the guard.

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -35,6 +35,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => address owner) public circleOwners;
 
   /// @dev Requires circle is commissioned by checking if an owner is set
   modifier onlyCommissioned(uint256 _id) {
@@ -88,6 +89,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
     isMember[_id][owner] = true;
     memberCircles[owner].push(_id);
     circleMembers[_id].push(owner);
+    circleOwners[_id] = owner;
 
     circles[_id] = _circle;
     emit CircleCreated(_id, _circle.token, _circle.depositAmount, _circle.depositInterval);
@@ -153,7 +155,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
       }
     }
 
-    delete circles[_id];
+    circles[_id].owner = address(0);
     emit CircleDecommissioned(_id);
   }
 
--- a/src/contracts/SavingCirclesViewer.sol
+++ b/src/contracts/SavingCirclesViewer.sol
@@ -130,7 +130,7 @@ contract SavingCirclesViewer is ISavingCirclesViewer {
 
     uint256 currentIndex = 0;
     for (uint256 i = 0; i < SAVING_CIRCLES.nextId(); i++) {
-      if (SAVING_CIRCLES.getCircle(i).owner == _user && !_isInArray(i, memberCircleIds)) {
+      if (_getCircleOwner(i) == _user && !_isInArray(i, memberCircleIds)) {
         ownedOnlyIds[currentIndex] = i;
         currentIndex++;
       }
@@ -144,7 +144,7 @@ contract SavingCirclesViewer is ISavingCirclesViewer {
     uint256[] memory memberCircleIds
   ) internal view returns (uint256 count) {
     for (uint256 i = 0; i < SAVING_CIRCLES.nextId(); i++) {
-      if (SAVING_CIRCLES.getCircle(i).owner == _user && !_isInArray(i, memberCircleIds)) {
+      if (_getCircleOwner(i) == _user && !_isInArray(i, memberCircleIds)) {
         count++;
       }
     }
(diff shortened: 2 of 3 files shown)


src/contracts/SavingCircles.sol
```diff
@@ -35,6 +35,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
   mapping(uint256 id => mapping(uint256 nonce => bool used)) public usedNonces;
   mapping(uint256 id => bool active) public isActive;
   mapping(uint256 id => address[] members) public circleMembers;
+  mapping(uint256 id => address owner) public circleOwners;
```
RonTuretzky, line 38: wouldn't a better design be to designate the 0 index to always be the circle owner?

src/contracts/SavingCirclesViewer.sol
```diff
@@ -265,14 +265,19 @@ contract SavingCirclesViewer is ISavingCirclesViewer {
...
     circleData.circleId = _circleId;
 
-    ISavingCircles.Circle memory circle = SAVING_CIRCLES.getCircle(_circleId);
+    ISavingCircles.Circle memory circle = _getCircle(_circleId);
+    bool isDecommissioned = SAVING_CIRCLES.isDecommissioned(circle);
 
-    if (SAVING_CIRCLES.isDecommissioned(circle)) {
+    if (circle.owner == address(0)) {
+      circle.owner = SAVING_CIRCLES.circleOwners(_circleId);
+    }
```
RonTuretzky, line 273: why would this ever happen?

RonTuretzky (changes requested):

RonTuretzky:
general note: this feels like a nice to have and outside of what i would slot into the current release

src/contracts/SavingCircles.sol
exo404: address(0) is equal to "Decommissioned" or "Stack doesn't exist" in our current semantic. What do you mean by 0 index. If you are pointing to the index in the members array, the owner is already the first in the array

src/contracts/SavingCirclesViewer.sol
exo404: he's updating the local circle variable with the owner after decommissioning the circle, but this approach feels messy imo

src/contracts/SavingCircles.sol
RonTuretzky: then why do we have this mapping? can't we just use that index instead of defining another variable ?

exo404: Yes I agree, also in the other PR

franrolotti: Good point. I added `circleOwners` as a way to preserve historical owner after `decommission()` zeroes `circle.owner`, so the viewer could still display owner data for decommissioned circles.

But I agree the cleaner design is to use the existing `circleMembers[id][0]`.

I’ll update this PR to remove the viewer dependency on `circleOwners` and reconstruct historical owner from `circleMembers[id][0]` instead.

src/contracts/SavingCirclesViewer.sol
franrolotti: It happens after decommission by design: `decommission()` sets `circles[id].owner = address(0)` to mark the circle as decommissioned.

That `if` was my patch to restore historical owner in viewer responses. I agree it’s messy with `circleOwners`; I’ll switch this to derive owner from `circleMembers[id][0]`.

If we want to move away from `owner == address(0)` as the decommission flag, I can open a follow-up PR for this or do it in this PR.

RonTuretzky: you can use circle end = type(uint256).max

RonTuretzky: i dont understand why we should need to restore a historical owner at any point so we should redesign if possible to avoid that

franrolotti: You’re right, and I agree.

I chose the `circleOwners`/viewer patch to avoid changing `decommission()` semantics in this PR, but that was probably the wrong tradeoff because it made everything messy and confusing.

I’ll switch to an explicit decommission marker in contract state and remove the owner-reconstruction logic from the viewer. Concretely:

1. On decommission, keep `owner` untouched.
2. Mark decommissioned via a dedicated rule (`circleEnd = type(uint256).max` as you suggested, or a dedicated `isDecommissionedById` flag).
3. Update `isDecommissioned`/guards to use that marker.
4. Remove `circleOwners` fallback logic from the viewer entirely.

Is this ok? @RonTuretzky

exo404: sounds good to me lmk if you need help on this

src/contracts/SavingCircles.sol
```diff
@@ -153,7 +155,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
       }
     }
 
-    delete circles[_id];
+    circles[_id].owner = address(0);
     emit CircleDecommissioned(_id);
```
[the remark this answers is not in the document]
exo404: @franrolotti seems important to note

src/contracts/SavingCircles.sol
```diff
@@ -88,6 +89,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuard, OwnableUpgradeable, E
     isMember[_id][owner] = true;
     memberCircles[owner].push(_id);
     circleMembers[_id].push(owner);
+    circleOwners[_id] = owner;
 
     circles[_id] = _circle;
```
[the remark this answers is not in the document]
exo404: no longer needed

src/contracts/SavingCirclesViewer.sol
RonTuretzky: sgtm

exo404:
@franrolotti @RonTuretzky @bagelface I'm fixing this in #144

PR #121, opened 2026-02-23
feat: add member-level withdrawability and fix viewer canWithdraw for late claims

franrolotti:
We need to change this in order for members to claim late because it fixes the `canWithdraw` from the viewer.

Today, late claims are valid at the contract level (`_claimable` allows a member to claim after their round has passed, as long as their round was fully funded and they have not already claimed), but the viewer only exposed withdrawability for the current round withdrawer. That makes `canWithdraw` incorrect for members who are still eligible to claim from an earlier round.

### What changed
- Added `isMemberWithdrawable(uint256 id, address member)` to `ISavingCircles` and `SavingCircles`
- Updated `SavingCirclesViewer` to compute `UserCircleData.canWithdraw` using `isMemberWithdrawable(circleId, user)` instead of `isCurrentWithdrawer && isWithdrawable(circleId)`
- Updated `isWithdrawable(uint256 id)` to return `true` if **any** member in the circle is claimable (aggregate circle-level status). 
- Updated fuzz test (`SavingCirclesMultiRound`) to use member-specific withdrawability when deciding whether the selected recipient should withdraw

Now:
- `viewer.canWithdraw` reflects whether the user can claim, including late claims
- `isWithdrawable` remains useful as a circle-level “someone can withdraw” signal

### Functions behavior 
- `isWithdrawable(id)`
  - Circle-level check
  - Returns `true` if the circle is active and **at least one member** is currently claimable
  - It no longer means “the current round withdrawer can withdraw”

- `isMemberWithdrawable(id, member)`
  - Member-level check
  - Returns `true` if the circle is active and that specific `member` is claimable right now

- `_claimable(id, member)`
  - Internal source of truth for claim eligibility
  - Returns `false` if the member already claimed or the circle is decommissionable
  - Returns `false` if the member is not in the circle or their turn has not been reached yet (`currentRound < memberIdx`)
  - Returns `true` only when all deposits for that member’s round are complete
  - This enables **late claims** because it allows `currentRound >= memberIdx` (not only `== memberIdx`)

### Test update note
The fuzz test now checks `isMemberWithdrawable(circleId, recipient)` before calling `withdraw`, which matches the actual recipient being tested and avoids relying on the broader circle-level `isWithdrawable` result.

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -296,7 +296,13 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     uint256 currentRound = _currentRoundIndex(_circle);
     if (currentRound >= circleMembers[_id].length) return false;
     address member = circleMembers[_id][currentRound];
-    return _claimable(_id, member);
+    return isMemberWithdrawable(_id, member);
+  }
+
+  /// @inheritdoc ISavingCircles
+  function isMemberWithdrawable(uint256 _id, address _member) public view override returns (bool) {
+    if (!isActive[_id]) return false;
+    return _claimable(_id, _member);
   }
 
   /// @inheritdoc ISavingCircles
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -296,7 +296,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
 
     address[] memory members = circleMembers[_id];
     for (uint256 i = 0; i < members.length; i++) {
-      if (_claimableWithoutDecommissionCheck(_id, members[i])) return true;
+      if (_activeClaimableCheck(_id, members[i])) return true;
     }
 
     return false;
@@ -391,10 +391,10 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
    */
   function _claimable(uint256 _id, address _member) internal view onlyCommissioned(_id) returns (bool) {
     if (_isDecommissionable(_id)) return false;
-    return _claimableWithoutDecommissionCheck(_id, _member);
+    return _activeClaimableCheck(_id, _member);
   }
 
-  function _claimableWithoutDecommissionCheck(uint256 _id, address _member) internal view returns (bool) {
+  function _activeClaimableCheck(uint256 _id, address _member) internal view returns (bool) {
     Circle memory _circle = circles[_id];
     if (_memberStates[_id][_member].hasClaimed) return false;
 
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -292,10 +292,11 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   /// @inheritdoc ISavingCircles
   function isWithdrawable(uint256 _id) public view override returns (bool) {
     if (!isActive[_id]) return false;
+    if (_isDecommissionable(_id)) return false;
 
     address[] memory members = circleMembers[_id];
     for (uint256 i = 0; i < members.length; i++) {
-      if (_claimable(_id, members[i])) return true;
+      if (_claimableWithoutDecommissionCheck(_id, members[i])) return true;
     }
 
     return false;
@@ -389,9 +390,13 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
    *      Claim eligibility is determined by the time-based round index.
    */
   function _claimable(uint256 _id, address _member) internal view onlyCommissioned(_id) returns (bool) {
+    if (_isDecommissionable(_id)) return false;
+    return _claimableWithoutDecommissionCheck(_id, _member);
+  }
+
+  function _claimableWithoutDecommissionCheck(uint256 _id, address _member) internal view returns (bool) {
     Circle memory _circle = circles[_id];
     if (_memberStates[_id][_member].hasClaimed) return false;
-    if (_isDecommissionable(_id)) return false;
 
     uint256 currentRound = _currentRoundIndex(_circle);
     (uint256 memberIdx, bool found) = _getMemberIndex(_id, _member);
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -292,11 +292,13 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   /// @inheritdoc ISavingCircles
   function isWithdrawable(uint256 _id) public view override returns (bool) {
     if (!isActive[_id]) return false;
-    Circle memory _circle = circles[_id];
-    uint256 currentRound = _currentRoundIndex(_circle);
-    if (currentRound >= circleMembers[_id].length) return false;
-    address member = circleMembers[_id][currentRound];
-    return isMemberWithdrawable(_id, member);
+
+    address[] memory members = circleMembers[_id];
+    for (uint256 i = 0; i < members.length; i++) {
+      if (_claimable(_id, members[i])) return true;
+    }
+
+    return false;
   }
 
   /// @inheritdoc ISavingCircles
--- a/src/interfaces/ISavingCircles.sol
+++ b/src/interfaces/ISavingCircles.sol
@@ -341,14 +341,14 @@ interface ISavingCircles {
   function isDecommissionable(uint256 id) external view returns (bool decommissionable);
 
   /**
-   * @notice Check if a circle is withdrawable
+   * @notice Check whether at least one member in a circle can currently claim a withdrawal
    * @param id The ID of the circle
-   * @return withdrawable Whether the circle is withdrawable
+   * @return withdrawable Whether any member in the circle is currently withdrawable
    */
   function isWithdrawable(uint256 id) external view returns (bool withdrawable);
 
   /**
-   * @notice Check if a specific member can withdraw for a circle
+   * @notice Check if a specific member is eligible to claim their withdrawal from a circle
    * @param id The ID of the circle
    * @param member The member to check
    * @return withdrawable Whether the member is withdrawable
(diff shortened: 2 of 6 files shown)


bagelface (changes requested):
## PR Review: feat: add member-level withdrawability and fix viewer canWithdraw for late claims

### Overall

The motivation is sound — `canWithdraw` in the viewer was incorrect for late-claiming members because it only checked `isCurrentWithdrawer && isWithdrawable`. The new `isMemberWithdrawable` cleanly solves this, and the viewer change to use it is correct. No security vulnerabilities found.

However, there are a few issues worth addressing before merge:

---

### 1. O(N²) gas in `isWithdrawable` — consider short-circuit optimization

`isWithdrawable` now loops all N members calling `_claimable` for each (SavingCircles.sol:297). Each `_claimable` call internally:
- Calls `_isDecommissionable(_id)` which itself loops all members via `_allMembersDepositedForRound` — O(N)
- Calls `_allMembersDepositedForRound` again — O(N)

This makes `isWithdrawable` **O(N²)**. While this is only a `view` function (free off-chain), it will cost real gas if called from another contract (e.g. a future integration, a keeper, or a composed DeFi protocol). The `_isDecommissionable` result is identical for every iteration of the loop and could be hoisted:

```solidity
function isWithdrawable(uint256 _id) public view override returns (bool) {
    if (!isActive[_id]) return false;
    if (_isDecommissionable(_id)) return false;

    address[] memory members = circleMembers[_id];
    for (uint256 i = 0; i < members.length; i++) {
        if (_claimableSkipDecommCheck(_id, members[i])) return true;
    }
    return false;
}
```

This would reduce worst-case to O(N) instead of O(N²). This isn't strictly necessary for correctness, but given this function is part of the public interface and may be called on-chain, it's worth the optimization.

### 2. Missing unit tests for `isMemberWithdrawable`

This PR adds a new public function to the interface but no dedicated tests. The following scenarios should have explicit coverage:

| Scenario | Expected |
|---|---|
| `isMemberWithdrawable` on a non-existent / inactive circle | `false` |
| `isMemberWithdrawable` for a non-member address | `false` (no revert) |
| `isMemberWithdrawable` for a member whose turn hasn't arrived | `false` |
| `isMemberWithdrawable` for a member who has already claimed | `false` |
| `isMemberWithdrawable` for a late-claimable member (past round, deposits complete) | `true` |
| `isMemberWithdrawable` when circle is decommissionable | `false` |

Several of these paths are exercised indirectly through existing `withdraw` tests, but since `isMemberWithdrawable` is now the source of truth for the viewer's `canWithdraw`, it deserves first-class test coverage.

### 3. `isWithdrawable` semantics are a breaking change — document it

The old `isWithdrawable` returned whether the **current-round withdrawer** could claim. The new version returns whether **any** member can claim. This is a different semantic. If any off-chain or on-chain consumer relied on the old meaning (e.g., a keeper that calls `isWithdrawable` then `withdraw` for the current-round member), they'd now get false positives when only a late-claimer is eligible but the current round isn't withdrawable.

Consider adding a `@notice` annotation on `isWithdrawable` clarifying the new aggregate semantics, and documenting the change in the interface natspec.

### 4. Fuzz test change is correct but doesn't expand late-claim coverage

The fuzz test update (SavingCirclesMultiRound.t.sol:532-538) correctly switches from `isWithdrawable` to `isMemberWithdrawable(circleId, recipient)`, which is more precise. However, `recipient` is still always `storedMembers[circleData.currentIndex]` — the current-round member. The fuzz suite never exercises the late-claim path (a past-round member claiming after their round passed). This is the primary new behavior enabled by this PR and would benefit from fuzz coverage.

### 5. Minor: `isMemberWithdrawable` natspec

The interface docstring says:
```
@notice Check if a specific member can withdraw for a circle
```

"withdraw for a circle" reads ambiguously — it could mean "withdraw on behalf of a circle" rather than "withdraw from a circle". Suggest:
```
@notice Check if a specific member is eligible to claim their withdrawal from a circle
```

---

### Verdict

The core logic changes are correct and the approach is clean. No vulnerabilities, no insolvency risks with the late-claim model (each claim always draws exactly one round's worth of deposits). The main requests are:

1. **Add unit tests for `isMemberWithdrawable`** (blocking)
2. **Consider the O(N²) optimization** (non-blocking but recommended)
3. **Update natspec for the `isWithdrawable` semantic change** (non-blocking)

franrolotti:
@bagelface [486c156](https://github.com/BreadchainCoop/saving-circles/pull/121/commits/486c156c8e1b482306e6d159632c2895a6867e4f) removes repeated decommission checks in `isWithdrawable`. For a full optimization, though, we’d need a small model change as tracking per-round funding completion in storage (e.g., funded-member counters or `roundFullyFunded` flags) and update it in `_deposit` when a member crosses from `< depositAmount` to `>= depositAmount`. Then claimability/decommissionability checks can use O(1) lookups instead of per-member scans. This is very feasible with current structures (`roundDeposits`, fixed member set after start), but we should do it in a focused follow-up PR to merge this one as it's needed for the current testing.

franrolotti:
## AI Review Summary

I reviewed this branch against `dev` and the changes are correct for the stated goal: fixing late-claim visibility in the viewer.

### What looks good
- `isMemberWithdrawable(id, member)` was added and wired correctly through `SavingCircles` and `ISavingCircles`.
- Viewer now computes `canWithdraw` using member-level eligibility, which fixes the late-claim mismatch.
- `isWithdrawable(id)` now reflects aggregate circle-level status (“someone can withdraw”), and the natspec/docs were updated to match that semantic.
- Unit coverage for `isMemberWithdrawable` is now solid (inactive/non-existent circle, non-member, future-round member, already-claimed member, late-claimable member, decommissionable circle).
- Viewer tests now cover late-claim behavior and decommissionable behavior correctly.
- The decommission check hoist in `isWithdrawable` is a good optimization step.

### Remaining notes
- `isWithdrawable` is still O(N²) worst-case because each member check still scans a round’s deposits. This is non-blocking here, but a storage-backed round-funded index/flag would be the right follow-up optimization.
- Fuzz update is directionally correct (`isMemberWithdrawable(circleId, recipient)`), but the helper still picks `recipient` as current-round member, so it does not fuzz true late-claimer actors yet.

### Verdict
No blocking issues found.  
The PR is mergeable for the intended fix, with performance/fuzz-depth improvements suitable for a focused follow-up PR.

src/contracts/SavingCircles.sol
```diff
@@ -292,11 +292,20 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
...
-    Circle memory _circle = circles[_id];
-    uint256 currentRound = _currentRoundIndex(_circle);
-    if (currentRound >= circleMembers[_id].length) return false;
-    address member = circleMembers[_id][currentRound];
-    return _claimable(_id, member);
+    if (_isDecommissionable(_id)) return false;
+
+    address[] memory members = circleMembers[_id];
+    for (uint256 i = 0; i < members.length; i++) {
+      if (_claimableWithoutDecommissionCheck(_id, members[i])) return true;
```
RonTuretzky: this function name is unnecessarly long

RonTuretzky: something like "activeClaimableCheck" maybe

src/interfaces/ISavingCircles.sol
```diff
@@ -341,12 +341,20 @@ interface ISavingCircles {
   function isDecommissionable(uint256 id) external view returns (bool decommissionable);
 
   /**
-   * @notice Check if a circle is withdrawable
+   * @notice Check whether at least one member in a circle can currently claim a withdrawal
```
RonTuretzky, line 344: I feel like this is misleading becuase it only checks whether the current round member can withdraw no?

franrolotti: It’s not current round-only: `isWithdrawable(id)` returns true if any member is currently claimable (including late claims), while `currentRoundWithdrawer(id)` is the current scheduled turn. This was a known limitation in https://github.com/BreadchainCoop/saving-circles/pull/107 to support late claim logic. Maybe we could revisit this and simplify/remove redundant functions (if any) in a follow-up.

Related to https://github.com/BreadchainCoop/saving-circles/pull/121#discussion_r2879067718

src/contracts/SavingCircles.sol
```diff
@@ -292,11 +292,20 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
...
+    }
+
+    return false;
```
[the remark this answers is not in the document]
franrolotti: Context on `isWithdrawable`: we kept/expanded it as a circle-level aggregate signal (“someone can withdraw now”), while `isMemberWithdrawable` is the source of truth for user-level eligibility (and what the viewer now uses for `canWithdraw`, including late claims).  

We could consider dropping/deprecating `isWithdrawable`, but may be a larger change for this PR (wdyt?). **We want this PR merged to correctly display the claim button for late claims in the current testing round.**

Then we could solve this, but not a blocker right now for our expected circle sizes (<10 members). Am I right here? @RonTuretzky

src/contracts/SavingCircles.sol
franrolotti: [c231e63](https://github.com/BreadchainCoop/saving-circles/pull/121/commits/c231e63e655c871e252f120643ffd2b6ecf5ef97)

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
bagelface: Agree on the time complexity not being a blocker, but please make an issue for it so we don't lose track of it.

bagelface (approved):
please open an issue (if there isn't already one) to optimize the time complexity of `isWithdrawable`. Other than that, lgtm

franrolotti:
@RonTuretzky do you think we could merge this? bagel already approved

src/contracts/SavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: agreed

RonTuretzky (approved):
Approving but please lets track the O(n^2) issue and I have a feeling that there are future viewer changes that need to be made to give the frontend everyone who's withdrawable at once

PR #137, opened 2026-03-31
Add automatic Gelato deposits

### LINKED ISSUE #136: Automatic Deposits
#118

exo404:
## Summary

This PR replaces the previous delegated deposit extension with a Gelato automatic deposit flow for Saving Circles.

The new implementation introduces `AutomaticSavingCircles`, a dedicated automation extension that lets users opt in once and then allows Gelato to execute deposits on their behalf when they are eligible. The final design no longer creates or resolves automation on a per-member or per-circle basis. Instead, Gelato calls a single global `checker()` that scans all circles and all members, and, when work is available, executes a single `executeAutomatedDeposits()`.

This aligns the product flow with the intended UX:
- a user enables automatic deposits once,
- Gelato monitors the system,
- Gelato deposits for every eligible opted-in member across every eligible circle.

## Motivation

The previous delegated deposit model exposed legacy manual execution paths and required more granular targeting than the intended automation flow.

This PR moves the feature to a stricter and clearer model:
- users explicitly opt in to automation,
- execution is restricted to a configured Gelato executor,
- the resolver/execution path is automation-oriented rather than manual,
- the naming is updated from `delegated` to `automatic` to reflect the actual product behavior.

### 1. Replaced the legacy delegated extension

The old `DelegatedSavingCircles` contract and interface were removed and replaced with:
- `AutomaticSavingCircles`
- `IAutomaticSavingCircles`

This also removes the old delegated terminology from the feature surface.

### 2. Added a Gelato-only execution model

The new extension is `Ownable` and stores an `automationExecutor` address, which is intended to be Gelato’s dedicated executor for the network.

Only that executor can trigger automated deposits:
- `setAutomationExecutor(address)` configures the authorized Gelato caller
- `executeAutomatedDeposits()` is protected by `onlyAutomationExecutor`

This ensures the automatic execution path is no longer generally permissionless.

### 3. Changed user opt-in from delegated deposits to automatic deposits

Users now opt in through:
- `setAutomaticDepositsEnabled(bool)`

The contract tracks this through:
- `automaticDepositsEnabled[address]`

This opt-in is global for the user, meaning once enabled, the user is considered for automation across every circle they belong to.

### 4. Reworked the Gelato resolver flow into a single global checker

The previous per-member style has been replaced by:
- `checker()`
- `executeAutomatedDeposits()`

`checker()` now:
- takes no arguments,
- scans every existing circle using `SavingCircles.nextId()`,
- returns `canExec = true` as soon as it finds at least one eligible automated deposit anywhere in the system,
- returns calldata for `executeAutomatedDeposits()`.

This means Gelato only needs to evaluate one resolver entrypoint rather than one task per member or one task per circle.

### 5. Added a global sweep execution path

`executeAutomatedDeposits()` performs a full automation sweep:
- iterates through every circle,
- skips circles that are missing, not started, expired, or decommissionable,
- loads member balances for each eligible circle,
- checks each member for opt-in status and deposit eligibility,
- completes deposits for every eligible member in the same transaction.

A member is considered eligible only if:
- they enabled automatic deposits,
- they still owe some amount for the current round,
- they have sufficient wallet balance,
- they have approved the automation contract for the missing amount.

If a circle or member is not eligible, the sweep skips them rather than reverting the whole execution.

### 6. Updated deployment wiring

The deployment helper now deploys `AutomaticSavingCircles` instead of `DelegatedSavingCircles` and passes the admin/owner into the constructor.

This keeps deployment aligned with the new ownership and executor-configuration model.

### 7. Rewrote the unit test suite

The old `DelegatedSavingCircles` tests were removed and replaced with `AutomaticSavingCircles` coverage.

The new tests cover:
- enabling and disabling automatic deposits,
- owner-only automation executor updates,
- Gelato-only execution restrictions,
- successful deposits across multiple circles in a single sweep,
- skipping ineligible members,
- skipping unstarted circles,
- partial deposit completion,
- global checker behavior when work exists and when it does not.

## Behavior After This PR

After this change, the expected lifecycle is:

1. A user joins one or more saving circles.
2. The user enables automatic deposits.
3. The user approves the `AutomaticSavingCircles` contract to spend the required token.
4. The owner configures the Gelato executor address.
5. Gelato calls `checker()`.
6. If any eligible automated deposit exists anywhere in the system, Gelato calls `executeAutomatedDeposits()`.
7. The contract deposits for every eligible opted-in member across every eligible circle in that execution.

## Why This Design

This design was chosen to support a simple automation model:
- one resolver entrypoint,
- one execution entrypoint,
- no need to create or maintain one task per user,
- no need to create or maintain one task per circle.

It gives a clearer operational model for automation and matches the product requirement that the system should automatically handle all opted-in users across all circles.

## Caveats / Follow-up

The main tradeoff of this design is execution size.

Because `checker()` and `executeAutomatedDeposits()` now scan globally, their cost scales with:
- the total number of circles
- the total number of members across those circles

This is acceptable for the current requested behavior, but if the system grows significantly, the next likely step would be to introduce pagination, cursors, or batched windows so automation remains reliable and cost-effective at larger scale.


RonTuretzky (commented):
instead of renaming ,deleting the file makes this pr very hard to review

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -0,0 +1,157 @@
...
+    if (!automaticDepositsEnabled[_member]) return false;
+    if (_currentBalance >= _circle.depositAmount) return false;
+
+    uint256 requiredAmount = _circle.depositAmount - _currentBalance;
+    IERC20 token = IERC20(_circle.token);
+
+    return token.allowance(_member, address(this)) >= requiredAmount && token.balanceOf(_member) >= requiredAmount;
+  }
+
+  function _executeAutomatedDepositsForCircle(uint256 _circleId) internal {
```
RonTuretzky: we should have a view function that returns all circles and all users in each circle that are eligbile for automation, instead of computing this on chain

RonTuretzky (changes requested):

exo404:
> instead of renaming ,deleting the file makes this pr very hard to review

I renamed the file, but for some reason when I rename a file the commit consider it as a delete

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -0,0 +1,157 @@
...
+        for (uint256 i = 0; i < members.length; i++) {
+          address member = members[i];
+          uint256 currentBalance = balances[i];
+          if (!_isEligibleForAutomatedDeposit(_circle, member, currentBalance)) continue;
+
+          uint256 requiredAmount = _circle.depositAmount - currentBalance;
+
+          token.safeTransferFrom(member, address(this), requiredAmount);
+          token.forceApprove(address(SAVING_CIRCLES), requiredAmount);
+          SAVING_CIRCLES.depositFor(_circleId, requiredAmount, member);
```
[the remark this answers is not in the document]
exo404: they cannot fail right? We check the conditions in the `isEligibleForAutomatedDeposit` function

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -0,0 +1,157 @@
...
+        }
+      } catch {}
+    } catch {}
```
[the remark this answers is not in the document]
exo404: [4fc57e5](https://github.com/BreadchainCoop/saving-circles/pull/137/commits/4fc57e56f3318aed0a04dec0ac47257cab150f12)

RonTuretzky:
> > instead of renaming ,deleting the file makes this pr very hard to review
> 
> I renamed the file, but for some reason when I rename a file the commit consider it as a delete

You have to use the command “git mv”

src/contracts/AutomaticSavingCircles.sol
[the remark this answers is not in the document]
RonTuretzky: Yea I agree this cannot fail

exo404:
> > > instead of renaming ,deleting the file makes this pr very hard to review
> > 
> > 
> > I renamed the file, but for some reason when I rename a file the commit consider it as a delete
> 
> You have to use the command “git mv”

got it thanks

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -0,0 +1,294 @@
...
+
+    return token.allowance(_member, address(this)) >= requiredAmount && token.balanceOf(_member) >= requiredAmount;
+  }
+
+  /**
+   * @dev External trampoline used to isolate single-target failures with try/catch during batch execution
+   * @param _circleId Circle to process
+   * @param _member Member to deposit for
+   */
+  function executeAutomatedDepositTarget(uint256 _circleId, address _member) external onlySelf {
```
bagelface: Confused by this. why not just call `_executeAutomatedDepositTarget` directly in `batchExecuteAutomatedDeposits`?

bagelface:
## Code Review: PR #137 — Add Automatic Gelato Deposits

**Overall:** This is a clean refactor with a solid security model. The trampoline + try/catch batch pattern is well-executed. A few issues below, ranging from bugs to nits.

---

### Issues

#### 1. Inaccurate error revert in `_executeAutomatedDepositTarget` (medium)

`AutomaticSavingCircles.sol:208–213`

```solidity
if (!_isCircleEligibleForAutomation(_circleId, _circle, members.length)) {
  if (!SAVING_CIRCLES.isActive(_circleId)) revert ISavingCircles.NotActive();
  if (block.timestamp < _circle.effectiveCircleStartTime) revert ISavingCircles.DepositBeforeCircleStart();
  if (SAVING_CIRCLES.isDecommissionable(_circleId)) revert ISavingCircles.CircleTimedOut();
  revert ISavingCircles.CircleExpired();
}
```

`_isCircleEligibleForAutomation` returns `false` when `effectiveCircleStartTime == 0` (circle not yet started/configured). In that case, the error path evaluates:
- `!isActive` → possibly false (circle can be active but unstarted)
- `block.timestamp < 0` → always false
- `isDecommissionable` → probably false

Falls through to `CircleExpired`, which is semantically wrong — the circle hasn't expired, it was never started. This is inside a try/catch batch loop so it doesn't break anything, but it produces misleading failure events (`AutomatedDepositFailed` with wrong reason). Adding an `effectiveCircleStartTime == 0` branch or a dedicated `CircleNotStarted` error would fix this.

---

#### 2. `checker()` description vs. implementation mismatch (low, but confusing)

The PR description says:
> `checker()` returns `canExec = true` **as soon as** it finds at least one eligible automated deposit

But the actual implementation calls `getEligibleAutomatedDeposits()`, which does a **full scan** of every circle and every member before returning. There is no early exit. The execPayload encodes **all** eligible deposits, not just the signal that one exists.

This is the better behavior (Gelato needs the full list), but the description is wrong. This could mislead whoever reads the PR later when reasoning about gas or scaling.

---

#### 3. Double full scan in `getEligibleAutomatedDeposits` (low)

`AutomaticSavingCircles.sol:121–135`

The function iterates all circles twice — once in `_countEligibleAutomatedDepositsForCircle` and once in `_populateEligibleAutomatedDepositsForCircle`. Each pass calls `SAVING_CIRCLES.getMemberBalances()`. Since `checker()` is a view function used off-chain, this doesn't affect transaction cost, but as the system grows this could affect Gelato's off-chain simulation performance. A single-pass approach using a dynamic array (or pre-allocated max-size array trimmed at the end) would fix this.

---

#### 4. `executeAutomatedDepositTarget` is missing a test for external access (low)

`AutomaticSavingCircles.sol:195–197`

The `onlySelf` modifier on this public trampoline function is the key safety guard preventing arbitrary external callers from triggering deposits. There is no test covering:

```solidity
vm.prank(alice);
vm.expectRevert(AutomaticSavingCircles.OnlySelf.selector);
automaticSavingCircles.executeAutomatedDepositTarget(circleId, alice);
```

Given this is a publicly-callable function on a contract that holds token approvals, this case should be covered.

---

#### 5. Deployment script doesn't capture the contract address (nit)

`script/Common.sol:43`

```solidity
new AutomaticSavingCircles(address(proxy), _admin);
```

The deployed address is discarded. Unless the address is derivable from CREATE2, there's no on-chain trace of where it was deployed. Since the executor address needs to be configured post-deploy via `setAutomationExecutor`, the deployer needs to know where the contract landed. Consider storing and logging the address.

---

#### 6. PR description refers to `executeAutomatedDeposits()` but function is `batchExecuteAutomatedDeposits()` (nit)

The "Behavior After This PR" section says Gelato calls `executeAutomatedDeposits()`. Actual function name is `batchExecuteAutomatedDeposits`. Minor doc inconsistency.

---

### What's Good

- **Security model is solid.** Even a compromised Gelato executor can only deposit for members who explicitly opted in AND granted allowance. The validation in `_executeAutomatedDepositTarget` re-checks all constraints on-chain, not just in the off-chain checker.
- **Try/catch trampoline pattern** (`batchExecuteAutomatedDeposits → this.executeAutomatedDepositTarget`) is the right approach for batch resilience. One failed member doesn't nuke the whole sweep.
- **`nonReentrant` placement** on `batchExecuteAutomatedDeposits` is correct — the self-call trampoline doesn't hit it, but external re-entry into batch execution is blocked.
- **`_isEligibleForAutomatedDeposit` matches `_executeAutomatedDepositTarget` constraints** — both check balance and allowance, so the checker payload accurately reflects what will succeed on execution.
- Test coverage is good overall — partial deposits, multi-circle sweeps, decommissioned circle skipping are all covered.

bagelface (approved):
I would definitely fix (1) mentioned here https://github.com/BreadchainCoop/saving-circles/pull/137#issuecomment-4210206169 (and any other ones you're feeling up to)

And the function ordering comment from the CI linter: 

"src/contracts/AutomaticSavingCircles.sol:R110": "Function order is incorrect, external view function can not go after public view function (line 87)"

But other than that this looks good to me.

src/contracts/AutomaticSavingCircles.sol
exo404: It's an internal function so I can't call it using the try/catch trampoline pattern right?

exo404: Also that could be useful to execute one target only I think

PR #145, opened 2026-04-04
refactor: use currentIndex sentinel instead of isDecommissioned mapping

RonTuretzky:
## Summary
- Replaces the `isDecommissioned` boolean mapping with a sentinel value (`type(uint256).max`) on the existing `currentIndex` field of the Circle struct
- Avoids adding new storage while preserving identical external behavior — `isDecommissioned(uint256)` remains callable as a view function
- All 141 tests pass

## Changes
- Removed `mapping(uint256 id => bool status) isDecommissioned`
- Added `DECOMMISSIONED_SENTINEL` constant (`type(uint256).max`)
- On decommission, sets `circles[_id].currentIndex = DECOMMISSIONED_SENTINEL` instead of `isDecommissioned[_id] = true`
- Added `_isDecommissionedCheck()` internal helper and `isDecommissioned()` public view function
- `getCircle()` still returns the computed round index (not the sentinel) since `_currentRoundIndex()` doesn't read `currentIndex`

## Test plan
- [x] All existing tests pass (141/141)
- [x] Build succeeds

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -44,11 +44,12 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   mapping(uint256 id => address[] members) public circleMembers;
   mapping(uint256 id => mapping(address member => MemberState state)) internal _memberStates;
   mapping(uint256 id => mapping(uint256 round => mapping(address member => uint256 amount))) public roundDeposits;
-  mapping(uint256 id => bool status) public override isDecommissioned;
+
+  uint256 public constant DECOMMISSIONED_SENTINEL = type(uint256).max;
 
   /// @dev Requires circle exists and has not been decommissioned
   modifier onlyCommissioned(uint256 _id) {
-    if (!_exists(_id) || isDecommissioned[_id]) revert NotCommissioned();
+    if (!_exists(_id) || _isDecommissionedCheck(_id)) revert NotCommissioned();
     _;
   }
 
@@ -161,7 +162,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     uint256 membersLength = members.length;
 
     isActive[_id] = false;
-    isDecommissioned[_id] = true;
+    circles[_id].currentIndex = DECOMMISSIONED_SENTINEL;
 
     // Return all funds still held by the contract to the members who deposited them.
     for (uint256 r = 0; r < membersLength; r++) {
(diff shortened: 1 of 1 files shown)


exo404:
this doesn't work @RonTuretzky because of the `_getCircle` function that overwrites the currentIndex field of the Circle struct. I still think that adding a mapping is better than messing with fields semantic.

src/contracts/SavingCircles.sol
exo404, line 219: look at this

exo404 (changes requested):

src/contracts/SavingCircles.sol
RonTuretzky: we can just change the _currentRoundIndex() method to not update it if it is the sentinel value?

RonTuretzky:
@bagelface @hudsonhrh @rubydusa we need a tie breaker here we spent too much time arguing about this lol

hudsonhrh:
> @bagelface @hudsonhrh @rubydusa we need a tie breaker here we spent too much time arguing about this lol

whats the argument exactly whether to use the mapping or not? usually I prefer max value for something like this because no new storage slot (saves gas) and is still o(1) check as long as everything is properly accounted for and edge cases handled. I agree mapping is easier and more intuitive but this is solidity lmao

rubydusa:
> @bagelface @hudsonhrh @rubydusa we need a tie breaker here we spent too much time arguing about this lol 

What's the purpose? To save a bit of gas?

RonTuretzky:
> > @bagelface @hudsonhrh @rubydusa we need a tie breaker here we spent too much time arguing about this lol
> 
> What's the purpose? To save a bit of gas?

save gas/ not add another variable

rubydusa:
> > > @bagelface @hudsonhrh @rubydusa we need a tie breaker here we spent too much time arguing about this lol
> > 
> > What's the purpose? To save a bit of gas?
> 
> save gas/ not add another variable 

Your change is fine imo

bagelface:
If it were me, I'd probably personally add the mapping since I see we have a bunch of other mappings already and I think it fits the best with the current design of the contract. It's more readability/straight forward. But no doubt this will save a cold SSTORE.

RonTuretzky:
> If it were me, I'd probably personally add the mapping since I see we have a bunch of other mappings already and I think it fits the best with the current design of the contract. It's more readability/straight forward. But no doubt this will save a cold SSTORE.

 OK, I think this settles it. I actually don’t care that much about the one called SS store Alberto if you think that this is the better way to do it let’s just do it

PR #157, opened 2026-04-13
Automatic claims

### LINKED ISSUE #141: Automatic Claiming
The round is distributed automatically when all members have deposited and the round's time condition is met.

🤖 Gelato trigger: Resolver — Gelato checks two conditions before executing: (1) all members have deposited, AND (2) the round time has passed. Both must be true simultaneously, which a pure time-based trigger can't guarantee alone.

exo404:
## Summary

This PR adds Gelato-powered automatic claims to the existing `AutomaticSavingCircles` automation extension.

Gelato can now resolve and execute claimable payouts automatically once both required conditions are true:
- all members have deposited for the payout round,
- the payout round’s time window has passed.

## What Changed

Added a new automatic claim flow:
- `claimChecker()` scans all circles and returns executable calldata when any member is eligible to claim.
- `getEligibleAutomatedClaims()` returns all eligible `(circleId, member)` claim targets.
- `batchExecuteAutomatedClaims(...)` executes claim targets through Gelato’s configured executor.
- `executeAutomatedClaimTarget(...)` is used internally so failed targets do not block the full batch with the try/catch pattern.
- `AutomatedClaimFailed(...)` is emitted when an individual claim target fails during batch execution.

Claim eligibility requires:
- the circle is active,
- the circle is not decommissionable,
- the member’s payout round time has passed,
- `SavingCircles.isMemberWithdrawable(circleId, member)` returns true.

This means Gelato only claims when deposits are complete and the time condition is satisfied.

## SavingCircles Change

`withdrawFor(...)` is now truly permissionless for valid claim recipients. The caller no longer needs to be a circle member; instead, the target member must be a valid member and must be withdrawable.

This is required because the automation contract is not a member of each circle, but it needs to trigger claims on behalf of eligible members. The payout safety checks still remain in place through `_claimable(...)`.

## Tests

Added coverage for:
- claim targets are not returned before the round time has passed,
- claim targets are not returned when deposits are incomplete,
- `claimChecker()` returns the correct batch execution payload,
- automatic claims execute across multiple circles,
- only the configured Gelato executor can run claim batches,
- mismatched claim batch arrays revert,
- non-member callers can trigger `withdrawFor(...)` for valid recipients,
- invalid claim recipients still revert.

Validation:
- `forge test`
Full suite passes: 156 tests.

### DIFF
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -118,32 +118,22 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
   /// @inheritdoc IAutomaticSavingCircles
   function depositChecker() external view override returns (bool canExec, bytes memory execPayload) {
-    uint256[] memory circleIds;
-    address[] memory members;
+    if (automationExecutor == address(0)) return (false, bytes(''));
 
-    if (automationExecutor != address(0)) {
-      (circleIds, members) = getEligibleAutomatedDeposits();
-      canExec = circleIds.length > 0;
-    } else {
-      circleIds = new uint256[](0);
-      members = new address[](0);
-    }
+    (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedDeposits();
+    canExec = circleIds.length > 0;
+    if (!canExec) return (false, bytes(''));
 
     execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedDeposits, (circleIds, members));
   }
 
   /// @inheritdoc IAutomaticSavingCircles
   function claimChecker() external view override returns (bool canExec, bytes memory execPayload) {
-    uint256[] memory circleIds;
-    address[] memory members;
+    if (automationExecutor == address(0)) return (false, bytes(''));
 
-    if (automationExecutor != address(0)) {
-      (circleIds, members) = getEligibleAutomatedClaims();
-      canExec = circleIds.length > 0;
-    } else {
-      circleIds = new uint256[](0);
-      members = new address[](0);
-    }
+    (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedClaims();
+    canExec = circleIds.length > 0;
+    if (!canExec) return (false, bytes(''));
 
     execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedClaims, (circleIds, members));
   }
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -118,10 +118,8 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
   /// @inheritdoc IAutomaticSavingCircles
   function depositChecker() external view override returns (bool canExec, bytes memory execPayload) {
-    if (automationExecutor == address(0)) return (false, bytes(''));
-
     (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedDeposits();
-    canExec = circleIds.length > 0;
+    canExec = _automationCanExecute(circleIds);
     if (!canExec) return (false, bytes(''));
 
     execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedDeposits, (circleIds, members));
@@ -129,10 +127,8 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
   /// @inheritdoc IAutomaticSavingCircles
   function claimChecker() external view override returns (bool canExec, bytes memory execPayload) {
-    if (automationExecutor == address(0)) return (false, bytes(''));
-
     (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedClaims();
-    canExec = circleIds.length > 0;
+    canExec = _automationCanExecute(circleIds);
     if (!canExec) return (false, bytes(''));
 
     execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedClaims, (circleIds, members));
@@ -182,6 +178,15 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     for (uint256 circleId = 0; circleId < circleCount; circleId++) {
       index = _appendEligibleAutomatedClaimsForCircle(circleId, circleIds, members, index);
     }
+  }
+
+  /**
+   * @dev Returns whether Gelato automation can execute for the selected targets
+   * @param _circleIds Circle IDs selected for automated execution
+   * @return Whether an automation executor is configured and at least one target exists
+   */
+  function _automationCanExecute(uint256[] memory _circleIds) internal view returns (bool) {
+    return automationExecutor != address(0) && _circleIds.length > 0;
   }
 
   /**
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -190,7 +190,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
     uint256 index = 0;
     for (uint256 circleId = 0; circleId < circleCount; circleId++) {
-      index = _populateEligibleAutomatedClaimsForCircle(circleId, circleIds, members, index);
+      index = _appendEligibleAutomatedClaimsForCircle(circleId, circleIds, members, index);
     }
   }
 
@@ -374,7 +374,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
    * @param _index Current write index in the output arrays
    * @return nextIndex Updated write index after appending any eligible targets
    */
-  function _populateEligibleAutomatedClaimsForCircle(
+  function _appendEligibleAutomatedClaimsForCircle(
     uint256 _circleId,
     uint256[] memory _circleIds,
     address[] memory _members,
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -72,6 +72,15 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     _executeAutomatedDepositTarget(_circleId, _member);
   }
 
+  /**
+   * @dev External trampoline used to isolate single-target failures with try/catch during batch claim execution
+   * @param _circleId Circle to process
+   * @param _member Member to claim for
+   */
+  function executeAutomatedClaimTarget(uint256 _circleId, address _member) external onlySelf {
+    _executeAutomatedClaimTarget(_circleId, _member);
+  }
+
   /// @inheritdoc IAutomaticSavingCircles
   function batchExecuteAutomatedDeposits(
     uint256[] calldata _circleIds,
@@ -88,12 +97,27 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
   }
 
   /// @inheritdoc IAutomaticSavingCircles
+  function batchExecuteAutomatedClaims(
+    uint256[] calldata _circleIds,
+    address[] calldata _members
+  ) external override nonReentrant onlyAutomationExecutor {
+    if (_circleIds.length != _members.length) revert ArrayLengthMismatch();
+
+    for (uint256 i = 0; i < _circleIds.length; i++) {
+      try this.executeAutomatedClaimTarget(_circleIds[i], _members[i]) {}
+      catch (bytes memory reason) {
+        emit AutomatedClaimFailed(_circleIds[i], _members[i], reason);
+      }
+    }
+  }
+
+  /// @inheritdoc IAutomaticSavingCircles
   function isAutomaticDepositsEnabled(address _member) external view override returns (bool) {
     return automaticDepositsEnabled[_member];
   }
 
   /// @inheritdoc IAutomaticSavingCircles
-  function checker() external view override returns (bool canExec, bytes memory execPayload) {
+  function depositChecker() external view override returns (bool canExec, bytes memory execPayload) {
     uint256[] memory circleIds;
     address[] memory members;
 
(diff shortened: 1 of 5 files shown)


src/interfaces/IAutomaticSavingCircles.sol
```diff
@@ -93,10 +108,24 @@ interface IAutomaticSavingCircles {
...
    * @notice Gelato resolver-style checker for automatic deposits across every circle
    * @return canExec Whether Gelato should execute the sweep
    * @return execPayload Encoded calldata for the automated deposit execution
    */
-  function checker() external view returns (bool canExec, bytes memory execPayload);
+  function depositChecker() external view returns (bool canExec, bytes memory execPayload);
+
+  /**
+   * @notice Gelato resolver-style checker for automatic claims across every circle
+   * @return canExec Whether Gelato should execute the claims
```
bagelface: There is a discrepancy here. need to either update this comment to "Whether Gelato can execute the claims" or change the name of the variable.

test/unit/AutomaticSavingCircles.t.sol
```diff
@@ -331,6 +337,103 @@ contract AutomaticSavingCirclesUnit is SavingCirclesTestBase {
...
+    vm.warp(circle.effectiveCircleStartTime + DEPOSIT_INTERVAL);
+
+    (bool canExec, bytes memory execPayload) = automaticSavingCircles.claimChecker();
+    assertTrue(canExec);
+
+    vm.prank(gelatoExecutor);
+    (bool success,) = address(automaticSavingCircles).call(execPayload);
+    assertTrue(success);
+
+    assertTrue(savingCircles.hasClaimed(baseCircleId, alice));
```
bagelface, line 395: check these values are false and balance is correct before the automated claim? maybe not needed since we dont seem to be checking pre-state for other tests

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -254,6 +366,39 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
 
+  /**
+   * @dev Appends the eligible claim targets for a circle into the output arrays used by the selector
+   * @param _circleId Circle to inspect
+   * @param _circleIds Output array of eligible circle IDs
+   * @param _members Output array of eligible members
+   * @param _index Current write index in the output arrays
+   * @return nextIndex Updated write index after appending any eligible targets
+   */
+  function _populateEligibleAutomatedClaimsForCircle(
```
bagelface: Naming of this function is a bit confusing, since it implies it's populating an array but really it's just returning the next index. The comments and function name prescribe a use to the function, which I think isn't ideal. Consider renaming to something like "_getNextEligibleAutomationClaimIndexForCircle"

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -163,6 +226,24 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
+    ISavingCircles.Circle memory _circle = SAVING_CIRCLES.getCircle(_circleId);
+    address[] memory members = SAVING_CIRCLES.getCircleMembers(_circleId);
+    (bool isMember, uint256 memberIndex) = _getMemberIndex(members, _member);
+
+    if (!isMember) revert ISavingCircles.NotMember();
+    if (!_isEligibleForAutomatedClaim(_circleId, _circle, _member, memberIndex)) {
+      revert ISavingCircles.NotWithdrawable();
+    }
+
+    SAVING_CIRCLES.withdrawFor(_circleId, _member);
```
bagelface, line 239: `_executeAutomatedClaimTarget` fetches `getCircleMembers` and calls `_getMemberIndex` to check membership, then calls `_isEligibleForAutomatedClaim`, which calls `isMemberWithdrawable → _claimable → _activeClaimableCheck → _getMemberIndex` — so membership is verified twice with two separate `getCircleMembers` calls. Since `withdrawFor → _withdraw` has `onlyMember(_id, _member)` as a hard guard, the explicit `NotMember` revert at line 239 is also redundant. Consider dropping the pre-check and relying on `withdrawFor` to enforce it, or at least share the member list with `_isEligibleForAutomatedClaim` to avoid the duplicate fetch.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -275,6 +420,29 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
+  function _isEligibleForAutomatedClaim(
+    uint256 _circleId,
+    ISavingCircles.Circle memory _circle,
+    address _member,
+    uint256 _memberIndex
+  ) internal view returns (bool) {
+    if (!SAVING_CIRCLES.isActive(_circleId)) return false;
+    if (_circle.effectiveCircleStartTime == 0) return false;
+    if (SAVING_CIRCLES.isDecommissionable(_circleId)) return false;
+    if (block.timestamp < _roundEndTime(_circle, _memberIndex)) return false;
```
bagelface, line 436: This assumes `_memberIndex` (the position in `getCircleMembers()`) equals the member's payout round index. That invariant holds as long as member order in the array is the canonical payout order, but it is implicit. Worth adding a comment asserting this assumption so a future refactor of how member order is stored does not silently break claim eligibility.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -221,6 +312,27 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
 
+  /**
+   * @dev Counts how many members in a circle are currently eligible for automated claim
+   * @param _circleId Circle to inspect
+   * @return eligibleCount Number of eligible claim targets found
+   */
+  function _countEligibleAutomatedClaimsForCircle(uint256 _circleId) internal view returns (uint256 eligibleCount) {
+    try SAVING_CIRCLES.getCircle(_circleId) returns (ISavingCircles.Circle memory _circle) {
+      if (!SAVING_CIRCLES.isActive(_circleId)) return 0;
+      if (SAVING_CIRCLES.isDecommissionable(_circleId)) return 0;
```
bagelface, line 318: `isActive` and `isDecommissionable` are checked here (and again in `_populateEligibleAutomatedClaimsForCircle`) before calling `_isEligibleForAutomatedClaim`, which already checks both conditions (lines 438–440). The early-exit guards are redundant and could diverge from the authoritative check over time. Consider removing them from `_countEligibleAutomatedClaimsForCircle` and `_populateEligibleAutomatedClaimsForCircle` and letting `_isEligibleForAutomatedClaim` be the single gate.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -108,6 +132,22 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
+
+    if (automationExecutor != address(0)) {
+      (circleIds, members) = getEligibleAutomatedClaims();
+      canExec = circleIds.length > 0;
+    } else {
+      circleIds = new uint256[](0);
+      members = new address[](0);
+    }
+
+    execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedClaims, (circleIds, members));
```
bagelface, line 134: When `automationExecutor == address(0)`, `canExec` is `false` but `execPayload` is still a valid (empty-arrays) encoded call. This is safe (the call would revert on `onlyAutomationExecutor`), but Gelato's own resolver examples use `execPayload = ""` when `canExec = false`. Low priority, but worth aligning for consistency with Gelato conventions.

test/unit/AutomaticSavingCircles.t.sol
```diff
@@ -331,6 +337,103 @@ contract AutomaticSavingCirclesUnit is SavingCirclesTestBase {
...
+    address[] memory targetMembers = new address[](0);
+    circleIds[0] = baseCircleId;
+
+    vm.prank(gelatoExecutor);
+    vm.expectRevert(abi.encodeWithSelector(IAutomaticSavingCircles.ArrayLengthMismatch.selector));
+    automaticSavingCircles.batchExecuteAutomatedClaims(circleIds, targetMembers);
+  }
+
   function _fundEnableAndApprove(address _member, uint256 _amount) internal {
     token.mint(_member, _amount);
```
bagelface, line 468: Missing parity with the deposit side: there is no test that a failed individual claim target (e.g. already claimed, or round time not yet passed) emits `AutomatedClaimFailed` and does not block the rest of the batch. Worth adding to match the coverage pattern for deposits.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -72,6 +72,15 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     _executeAutomatedDepositTarget(_circleId, _member);
   }
 
+  /**
+   * @dev External trampoline used to isolate single-target failures with try/catch during batch claim execution
+   * @param _circleId Circle to process
+   * @param _member Member to claim for
+   */
+  function executeAutomatedClaimTarget(uint256 _circleId, address _member) external onlySelf {
```
bagelface, line 80: The rename of `checker()` to `depositChecker()` changes the function selector — a breaking change for any live Gelato resolver pointing at the old selector. Since `AutomaticSavingCircles` is not upgradeable, existing Gelato tasks will need to be manually reconfigured to point at both `depositChecker` and the new `claimChecker`. Worth confirming this is accounted for in the deployment plan.

bagelface (commented):
This and every comment in this thread was generated by Claude 🤖 

Overall the PR is well-structured and consistent with the existing deposit automation pattern. A few issues below, not covered by existing comments.

src/contracts/AutomaticSavingCircles.sol
bagelface: I think we make this assumption throughout the contract, so can probably skip this

src/contracts/AutomaticSavingCircles.sol
bagelface: Assuming we aren't live yet, this isn't important.

src/contracts/AutomaticSavingCircles.sol
bagelface: I generally agree with this comment, but assuming the early-exit guards aren't too expensive it can be worth leaving in just to future proof. `isDecommissionable` loops over members, so this is sort of a heavy check.

bagelface (changes requested):
Got a few comments on here that I think are worth addressing.

src/contracts/AutomaticSavingCircles.sol
franrolotti: wdyt about `_appendEligibleAutomatedClaimsForCircle`? since the helper does write into the output arrays and then returns the next index, that name feels a bit closer to what it’s doing

@bagelface

src/contracts/AutomaticSavingCircles.sol
franrolotti: yeah, fair. We still need the member index for the round time check, so I don’t think I can just drop the lookup without changing the invalid member path. I can either keep the pre-check, or refactor this so membership/index is resolved once and reused more cleanly. wdyt? @bagelface

src/contracts/AutomaticSavingCircles.sol
franrolotti: skipping this one @bagelface

src/contracts/AutomaticSavingCircles.sol
franrolotti: skipping this one too

src/contracts/AutomaticSavingCircles.sol
franrolotti: Yeah makes sense. Since this is safe and both checkers currently use the same empty batch payload pattern, we could leave as is. for this PR unless you’d prefer we align both with Gelato’s `bytes("")` convention

@bagelface

src/interfaces/IAutomaticSavingCircles.sol
franrolotti: [0bfcded](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/0bfcdedb00ea948e54d7be31af8d29e052127c92)

test/unit/AutomaticSavingCircles.t.sol
franrolotti: [bacc792](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/bacc79207894810f5ef7c843b0e22df53e79d1fd)

test/unit/AutomaticSavingCircles.t.sol
franrolotti: [24ae60c](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/24ae60c5c34a40aff8b2a29ddf81b3465dd7fefe)

src/contracts/AutomaticSavingCircles.sol
bagelface: yea that sounds good

src/contracts/AutomaticSavingCircles.sol
bagelface: I would recommend just fixing it now since it's pretty low lift and the scope of this PR is exactly this

src/contracts/AutomaticSavingCircles.sol
bagelface: I would advocate for refactor so membership/index is resolved only once, but will defer to you to make final call. At the very least it would be good to open an issue for it so we can fix it if we ever do gas optimization work on these contracts.

franrolotti: I will open an issue as I don't want to block the PR with possible back and forth comments. Because we are planning to launch tomorrow @bagelface

src/contracts/AutomaticSavingCircles.sol
franrolotti: [c981bc2](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/c981bc2a0fc72be1e522fb801f40a8c5bcb2ce2a)

src/contracts/AutomaticSavingCircles.sol
franrolotti: [3886c43](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/3886c438d1f33442a8e3c3c0cff50b80f4c19d49)

bagelface: Since it's checked in two places I would break this out into an internal function so there's no chance the checks diverge

```solidity
/// natspec
function _automationCanExecute(uint256[] memory _circleIds) internal view returns (bool) {
  return automationExecutor != address(0) && _circleIds.length > 0;
}
```

franrolotti: [d4c23f8](https://github.com/BreadchainCoop/saving-circles/pull/157/commits/d4c23f880c0fdbd5c98f5a349983e96c5b4c8dc2)

PR #161, opened 2026-04-24
Rebase Rubydusa branch

franrolotti:


.github/workflows/tests.yml
```diff
@@ -31,6 +31,110 @@ jobs:
...
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile --network-concurrency 1
+
+      - name: Precompile using 0.8.14 and via-ir=false
+        run: yarn build
+
+      - name: Run tests
+        shell: bash
+        run: yarn test:unit
```
rubydusa, line 59: Forge unit tests are covered by etherform: https://github.com/BreadchainCoop/etherform/blob/a566c75290437054eeb28a1a059e008e0a0a05e7/.github/workflows/_foundry-cicd.yml#L176

Is there a specific reason to run the yarn test script over forge test?

.github/workflows/tests.yml
```diff
@@ -31,6 +31,110 @@ jobs:
...
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile --network-concurrency 1
+
+      - name: Precompile using 0.8.14 and via-ir=false
+        run: yarn build
+
+      - name: Run tests
+        run: yarn test:integration
```
rubydusa, line 85: same comment as for unit tests

.github/workflows/tests.yml
```diff
@@ -31,6 +31,110 @@ jobs:
...
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile --network-concurrency 1
+
+      - name: Precompile using 0.8.14 and via-ir=false
+        run: yarn build
+
+      - name: Run fuzz tests
+        run: forge test --match-contract Fuzz -vvv
```
rubydusa, line 111: same comment as unit tests

.github/workflows/tests.yml
```diff
@@ -31,6 +31,110 @@ jobs:
...
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile --network-concurrency 1
+
+      - name: Precompile with via-ir=false
+        run: yarn build
+
+      - name: Run tests
+        run: yarn test:integration
```
rubydusa, line 137: this looks wrong: halmos tests runs yarn test:integration?

script/Common.sol
rubydusa, line 1: these changes seem unrelated to etherform integration - just to make sure, these are updating the etherform integration to a more updated branch?

foundry.toml
rubydusa, line 1: These changes seems to have flaked in here - they are not good since having in the profile default build info compilation slows things down significantly

I added to etherform anyways to build it with these flags itself such that it does not depend on the repo's build profiles

I reverted those changes in my branch (this is why it appears here as added back) so this is probably the result of bad merge conflict resolution

rubydusa (changes requested):
There is a lot of changes here that seem to be unrelated to etherform integration, I suspect something had been done weirdly during merge conflict resolution or with the whole rebasing

I re-checked my PR and there are no merge conflicts with dev so I don't see a reason for solidity file changes to be introduced to my branch

PR #172, opened 2026-05-26
Add per-circle opt-in for automatic claims

franrolotti:
## Summary

Adds per-circle user opt-in for automatic claims processed through `AutomaticSavingCircles`.

Users can now enable or disable automated claims independently for each saving circle, and Gelato will execute eligible opted-in claims in shared batches.

## Changes

- Added `setAutomaticClaimsEnabled(uint256 circleId, bool enabled)`.
- Added `isAutomaticClaimsEnabled(uint256 circleId, address member).`
- Added per-circle/member automatic claim preference storage.
- Added `AutomaticClaimsToggled` event.
- Added `AutomaticClaimsNotEnabled` error.
- Updated claim eligibility selection to include only opted-in `(circleId, member)` pairs.
- Revalidated opt-in during batch execution so executor-supplied targets cannot bypass user preference.
- Added unit tests for claim opt-in toggling, per-circle filtering, batch execution protection, and existing batched claim behavior.

## Scope

This PR changes automatic claims only.

Automatic deposits are outside MVP scope. Existing deposit opt-in remains member-wide across all eligible circles and must be revisited before deposit automation is exposed through the frontend or enabled in Gelato.

### DIFF
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -23,8 +23,8 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
   /// @notice Dedicated Gelato msg.sender allowed to execute automated batches
   address public automationExecutor;
 
-  /// @notice Mapping to track which members have enabled automatic deposits
-  mapping(address member => bool enabled) public automaticDepositsEnabled;
+  /// @notice Mapping to track which members have enabled automatic deposits in each circle
+  mapping(uint256 circleId => mapping(address member => bool enabled)) public automaticDepositsEnabled;
 
   /// @notice Mapping to track which members have enabled automatic claims in each circle
   mapping(uint256 circleId => mapping(address member => bool enabled)) public automaticClaimsEnabled;
@@ -54,9 +54,9 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
   }
 
   /// @inheritdoc IAutomaticSavingCircles
-  function setAutomaticDepositsEnabled(bool _enabled) external override {
-    automaticDepositsEnabled[msg.sender] = _enabled;
-    emit AutomaticDepositsToggled(msg.sender, _enabled);
+  function setAutomaticDepositsEnabled(uint256 _circleId, bool _enabled) external override {
+    automaticDepositsEnabled[_circleId][msg.sender] = _enabled;
+    emit AutomaticDepositsToggled(_circleId, msg.sender, _enabled);
   }
 
   /// @inheritdoc IAutomaticSavingCircles
@@ -121,8 +121,8 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
   }
 
   /// @inheritdoc IAutomaticSavingCircles
-  function isAutomaticDepositsEnabled(address _member) external view override returns (bool) {
-    return automaticDepositsEnabled[_member];
+  function isAutomaticDepositsEnabled(uint256 _circleId, address _member) external view override returns (bool) {
+    return automaticDepositsEnabled[_circleId][_member];
   }
 
   /// @inheritdoc IAutomaticSavingCircles
@@ -221,7 +221,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
     (bool isMember, uint256 currentBalance) = _getMemberBalance(members, balances, _member);
     if (!isMember) revert ISavingCircles.NotMember();
-    if (!automaticDepositsEnabled[_member]) revert AutomaticDepositsNotEnabled();
+    if (!automaticDepositsEnabled[_circleId][_member]) revert AutomaticDepositsNotEnabled();
     if (currentBalance >= _circle.depositAmount) revert ISavingCircles.AlreadyDeposited();
 
     uint256 requiredAmount = _circle.depositAmount - currentBalance;
@@ -284,17 +284,19 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
   /**
    * @dev Returns whether a member currently satisfies the offchain selection criteria for automated deposit
+   * @param _circleId Circle to inspect
    * @param _circle Circle configuration to evaluate
    * @param _member Member being checked
    * @param _currentBalance Amount already deposited by the member in the active round
    * @return Whether the member can be included in the automation payload
    */
   function _isEligibleForAutomatedDeposit(
+    uint256 _circleId,
     ISavingCircles.Circle memory _circle,
     address _member,
     uint256 _currentBalance
   ) internal view returns (bool) {
-    if (!automaticDepositsEnabled[_member]) return false;
+    if (!automaticDepositsEnabled[_circleId][_member]) return false;
     if (_currentBalance >= _circle.depositAmount) return false;
 
     uint256 requiredAmount = _circle.depositAmount - _currentBalance;
@@ -315,7 +317,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
       if (!_isCircleEligibleForAutomation(_circleId, _circle, members.length)) return 0;
 
       for (uint256 i = 0; i < members.length; i++) {
-        if (_isEligibleForAutomatedDeposit(_circle, members[i], balances[i])) {
+        if (_isEligibleForAutomatedDeposit(_circleId, _circle, members[i], balances[i])) {
           eligibleCount++;
         }
       }
@@ -367,7 +369,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
       if (!_isCircleEligibleForAutomation(_circleId, _circle, members.length)) return nextIndex;
 
       for (uint256 i = 0; i < members.length; i++) {
-        if (!_isEligibleForAutomatedDeposit(_circle, members[i], balances[i])) continue;
+        if (!_isEligibleForAutomatedDeposit(_circleId, _circle, members[i], balances[i])) continue;
 
         _circleIds[nextIndex] = _circleId;
         _members[nextIndex] = members[i];
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -246,6 +246,8 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     (bool isMember, uint256 memberIndex) = _getMemberIndex(members, _member);
 
     if (!isMember) revert ISavingCircles.NotMember();
+    // Re-checked here so executor-supplied targets get a specific AutomaticClaimsNotEnabled error
+    // instead of the generic NotWithdrawable returned by _isEligibleForAutomatedClaim.
     if (!automaticClaimsEnabled[_circleId][_member]) revert AutomaticClaimsNotEnabled();
     if (!_isEligibleForAutomatedClaim(_circleId, _circle, _member, memberIndex)) {
       revert ISavingCircles.NotWithdrawable();
--- a/src/contracts/AutomaticSavingCircles.sol
+++ b/src/contracts/AutomaticSavingCircles.sol
@@ -20,7 +20,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
   /// @notice The main SavingCircles contract
   ISavingCircles public immutable SAVING_CIRCLES;
 
-  /// @notice Dedicated Gelato msg.sender allowed to execute automated batches
+  /// @notice Dedicated Gelato `msg.sender` allowed to execute automated batches
   address public automationExecutor;
 
   /// @notice Mapping to track which members have enabled automatic deposits in each circle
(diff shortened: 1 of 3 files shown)


franrolotti (commented):
_(Superseded — see fresh review below.)_

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -232,6 +246,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     (bool isMember, uint256 memberIndex) = _getMemberIndex(members, _member);
 
     if (!isMember) revert ISavingCircles.NotMember();
+    if (!automaticClaimsEnabled[_circleId][_member]) revert AutomaticClaimsNotEnabled();
```
franrolotti: Suggestion: add a one-line comment explaining why this explicit check is kept even though `_isEligibleForAutomatedClaim` (line 448) already filters on `automaticClaimsEnabled`. Without context a future reader is likely to remove one as dead code.

```solidity
// Re-checked here so executor-supplied targets get a specific AutomaticClaimsNotEnabled error
// instead of the generic NotWithdrawable returned by _isEligibleForAutomatedClaim.
if (!automaticClaimsEnabled[_circleId][_member]) revert AutomaticClaimsNotEnabled();
```

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -13,19 +13,22 @@ using SafeERC20 for IERC20;
...
 contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyGuard {
   /// @notice The main SavingCircles contract
   ISavingCircles public immutable SAVING_CIRCLES;
 
-  /// @notice Dedicated Gelato msg.sender allowed to execute automated deposit batches
+  /// @notice Dedicated Gelato msg.sender allowed to execute automated batches
   address public automationExecutor;
 
   /// @notice Mapping to track which members have enabled automatic deposits
   mapping(address member => bool enabled) public automaticDepositsEnabled;
```
franrolotti: Worth dropping a `// TODO:` here noting that this is intentionally member-wide (vs. the new per-circle `automaticClaimsEnabled` on line 30) and must be revisited before deposit automation is exposed in the frontend / Gelato. The constraint is in the PR description but won't survive into the codebase.

src/interfaces/IAutomaticSavingCircles.sol
```diff
@@ -69,7 +82,14 @@ interface IAutomaticSavingCircles {
   function setAutomaticDepositsEnabled(bool enabled) external;
 
   /**
-   * @notice Configure the Gelato dedicated msg.sender allowed to execute automated deposits
+   * @notice Enable or disable automatic claims for the caller in one circle
```
franrolotti: Nit: *"for one circle"* reads more naturally than *"in one circle"*.

```suggestion
   * @notice Enable or disable automatic claims for the caller for one circle
```

src/interfaces/IAutomaticSavingCircles.sol
```diff
@@ -47,6 +55,11 @@ interface IAutomaticSavingCircles {
    */
   error AutomaticDepositsNotEnabled();
 
+  /**
+   * @notice Thrown when automatic claims have not been enabled for a member in a circle
+   */
+  error AutomaticClaimsNotEnabled();
```
franrolotti: Optional: consider parameterizing this error so an isolated revert decode (outside the batch `AutomatedClaimFailed` event) still carries enough context to debug:

```solidity
error AutomaticClaimsNotEnabled(uint256 circleId, address member);
```

Nice-to-have for direct/external callers — `AutomatedClaimFailed` already includes both in the batch path.

franrolotti (commented):
## AI review (Claude Opus 4.7)

Reviewed against the PR branch checked out locally; `forge test --match-path 'test/unit/AutomaticSavingCircles.t.sol'` passes 29/29. No deployment-script or integration-test impact (constructor signature unchanged).

Overall: clean, well-scoped change. Storage layout, event/error additions, and the double-check pattern (selector filter + executor trampoline re-validation) are correct, and the new tests cover toggle, per-circle selector filtering, and batch refusal without opt-in.

### Findings

1. **Redundant opt-in check on the claim execution path.** `_executeAutomatedClaimTarget` checks `automaticClaimsEnabled` explicitly (line 249) *and* `_isEligibleForAutomatedClaim` checks it again (line 448). Keeping both is intentional and useful — the explicit check surfaces `AutomaticClaimsNotEnabled` instead of the generic `NotWithdrawable` — but worth a one-line comment so a future reader doesn't delete one as dead code. (See inline.)

2. **Opt-in asymmetry between deposits and claims.** Deposits remain member-wide (`automaticDepositsEnabled[member]`) while claims are per-circle (`automaticClaimsEnabled[circleId][member]`). The PR description notes this is intentional for MVP and must be revisited before deposit automation ships. Worth a `// TODO:` near `automaticDepositsEnabled` so the constraint survives in code, not just PR history. (See inline.)

3. **Claim opt-in is sticky.** Once a member opts in for a circle, every future eligible round is claimed automatically until they explicitly toggle off. Likely intentional, but a user who only wants the *next* round automated has no "claim once" UX. Worth confirming whether that matches product intent.

4. **Error specificity asymmetry between deposit and claim execution paths.** `_executeAutomatedDepositTarget` (lines 215–220) distinguishes `NotActive` / `CircleTimedOut` / `DepositBeforeCircleStart` / `CircleNotStarted` so failures emitted via `AutomatedDepositFailed.reason` are directly actionable off-chain. `_executeAutomatedClaimTarget` (lines 250–252) collapses every non-opt-in failure into generic `ISavingCircles.NotWithdrawable`. Not a correctness issue, but operators debugging `AutomatedClaimFailed` get less context than on the deposit side. Optional follow-up: mirror the deposit-side error tree (round time not yet passed vs. deposits incomplete vs. already claimed, etc.).

5. **Stale constructor NatSpec.** `AutomaticSavingCircles.sol` line 50 still reads *"Owner of the automatic deposits extension"*. Pre-existing copy, but now stale given the title/notice NatSpec at the top of the file was updated to cover both deposits and claims. Suggest "Owner of the automation extension".

### Nits
- Interface NatSpec for `setAutomaticClaimsEnabled` says *"in one circle"* — *"for one circle"* reads more naturally. (See inline.)
- `AutomaticClaimsNotEnabled` has no parameters; consider `(uint256 circleId, address member)` for easier debugging when this error is decoded in isolation. `AutomatedClaimFailed` already carries both in the batch path, so this is optional. (See inline.)

### Verdict
LGTM. All findings are non-blocking polish.

---
*Generated by Claude Opus 4.7 via Claude Code*

test/unit/AutomaticSavingCircles.t.sol
```diff
@@ -473,6 +537,11 @@ contract AutomaticSavingCirclesUnit is SavingCirclesTestBase {
     token.approve(address(automaticSavingCircles), _amount);
   }
 
+  function _enableAutomaticClaim(uint256 _circleId, address _member) internal {
+    vm.prank(_member);
+    automaticSavingCircles.setAutomaticClaimsEnabled(_circleId, true);
```
bagelface, line 599: add a check of the `isAutomaticClaimsEnabled` flag before and after setting

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -13,19 +13,22 @@ using SafeERC20 for IERC20;
...
 
+  /// @notice Mapping to track which members have enabled automatic claims in each circle
+  mapping(uint256 circleId => mapping(address member => bool enabled)) public automaticClaimsEnabled;
```
bagelface, line 30: We should try to match the variable types here. Why do claim flags need to be a double mapping from `circleId`, but deposit flags don't?

src/interfaces/IAutomaticSavingCircles.sol
bagelface: `AutomaticClaimsDisabled` is a better name for this imo

src/interfaces/IAutomaticSavingCircles.sol
```diff
@@ -69,7 +82,14 @@ interface IAutomaticSavingCircles {
...
   /**
-   * @notice Configure the Gelato dedicated msg.sender allowed to execute automated deposits
+   * @notice Enable or disable automatic claims for the caller in one circle
+   * @param circleId The circle to configure
+   * @param enabled Whether to enable automatic claims
+   */
+  function setAutomaticClaimsEnabled(uint256 circleId, bool enabled) external;
+
+  /**
+   * @notice Configure the Gelato dedicated msg.sender allowed to execute automated batches
```
bagelface: Personal preference, but I like putting references to code elements in `` tags like `msg.sender`

src/contracts/AutomaticSavingCircles.sol
bagelface: I would move this to the top of the `_executeAutomatedClaimTarget` function so it fails without running through the `getCircleMembers` and `_getMemberIndex` functions.

bagelface (changes requested):
A few comments to address.

I do see how the discrepancy between `automaticClaimsEnabled` and `automaticDepositsEnabled` happened.. we want the user to be able to set automatic claims "per instance" while deposits are globally either on or off for all circles. I think that's a bit confusing tbh and it'd be better if both of these features worked the same way (either both are per-instance or both are global). It would likely require a bit of a refactor of the features though. Maybe for example redefining the data structure as:

`mapping(address member => uint256[] circleIds) public automaticDepositsEnabled;`
`mapping(address member => uint256[] circleIds) public automaticClaimsEnabled;`

or

`mapping(address member => mapping(uint256 circleId => bool enabled)) public automaticDepositsEnabled;`
`mapping(address member => mapping(uint256 circleId => bool enabled)) public automaticClaimsEnabled;`

Hard to say which is better off the top of my head. Anyways, just a thought, I don't want to block over this.

BreadrichEngels:
## 🤖 Breadbrich Engels review — new items only

Read all existing reviews before commenting. Adding two points not yet covered:

---

### 1. `setAutomaticClaimsEnabled` accepts any `circleId` — including non-existent or completed circles

There's no validation that `_circleId` refers to an existing circle, or that `msg.sender` is actually a member of it:

```solidity
function setAutomaticClaimsEnabled(uint256 _circleId, bool _enabled) external override {
    automaticClaimsEnabled[_circleId][msg.sender] = _enabled;
    emit AutomaticClaimsToggled(_circleId, msg.sender, _enabled);  // fires for any circleId
}
```

Functionally harmless — the eligibility checks downstream handle it — but off-chain indexers consuming `AutomaticClaimsToggled` events may surface phantom opt-ins to UIs for circles that don't exist or that the user was never part of. Worth at least a `@dev` NatSpec note that the function does not validate circle existence or membership, so consumers know to treat the event as a preference record only.

Compare: `setAutomaticDepositsEnabled` has the same pattern, so this is consistent — just noting it here as new surface area introduced by the per-circle keying.

---

### 2. Opt-in storage persists after a circle's lifecycle ends

`automaticClaimsEnabled[circleId][member]` is written on opt-in but never cleared when a circle completes, is decommissioned, or when a member exits. For now this is fine — IDs are monotonically increasing so the slot is effectively dead once the circle is done — but it does mean:

- A member who opts in, forgets about it, and later calls `isAutomaticClaimsEnabled(circleId, member)` on a finished circle gets `true` back, which is technically correct but potentially surprising.
- If a future refactor ever makes `circleId` reusable (e.g., a v2 factory pattern), the stale opt-in would silently carry over.

Not blocking, but worth a `@dev` note on `automaticClaimsEnabled` similar to the `automaticDepositsEnabled` TODO that was suggested in the other review.

---

Both are non-blocking polish, consistent with the existing LGTM from the earlier review. The architectural asymmetry (bagelface's main comment) is the one I'd address before merge.

src/interfaces/IAutomaticSavingCircles.sol
franrolotti: Done in ba8619e.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Done in ccdf96d — added the inline comment explaining the intentional duplication.

src/interfaces/IAutomaticSavingCircles.sol
franrolotti: Agreed, renamed in 39f60e5.

test/unit/AutomaticSavingCircles.t.sol
franrolotti: Done in 9e6a268 — the helper now asserts `isAutomaticClaimsEnabled` is false before the toggle and true after.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Done in 4ba7685 — `automaticDepositsEnabled` is now also a per-circle mapping, matching `automaticClaimsEnabled`.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Done in 30d9da7 — the opt-in check now runs before `getCircleMembers` / `_getMemberIndex`.

src/interfaces/IAutomaticSavingCircles.sol
franrolotti: Done in 9f2209a — `msg.sender` references in the automation NatSpec are now wrapped in backticks.

bagelface (approved):
My comments look fixed, so approving.

PR #173, opened 2026-05-28
Refactor automation from Gelato to Chainlink

franrolotti:
This PR moves automatic claim execution to the Chainlink Automation pattern directly inside `AutomaticSavingCircles`, replacing the previous Gelato-style resolver flow.

Reason for the move: Gelato’s current dashboard appears to be in transition and has inconsistencies with its older Automate/resolver documentation, which made the required task configuration difficult to find and validate. Chainlink Automation was tested end-to-end on Sepolia and works reliably with our claim flow.

Changes:

- Adds `checkUpkeep` and `performUpkeep` to `AutomaticSavingCircles`.
- Adds unit coverage for the direct Chainlink execution path


franrolotti:
## Sepolia Verification
Deployed contracts on Sepolia:

- `SavingCircles` implementation: `0xb5bfd59ef49eba6a22213c257fa5134f8a9017b7`
- `SavingCircles` proxy (TransparentUpgradeableProxy): `0x711dd11384539e76a8ceab00eedfaf017c1d86a3`
- `AutomaticSavingCircles`: `0xd9c4e9bc89a7139849fdccf5d4eda50370d90e8f`
- `SavingCirclesViewer`: `0x954431910baafb6a0cdfa2d6d951f2a4bd69bba6`

All four contracts verified on Etherscan.

Chainlink Automation wired.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -110,14 +110,25 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
     }
+
+    (uint256[] memory circleIds, address[] memory members) = abi.decode(_performData[4:], (uint256[], address[]));
+    _batchExecuteAutomatedClaims(circleIds, members);
+  }
+
+  /// @inheritdoc IAutomaticSavingCircles
+  function checkUpkeep(bytes calldata) external view override returns (bool upkeepNeeded, bytes memory performData) {
+    return _claimChecker();
   }
```
franrolotti, line 132: **Finding 1 — `cannotExecute` protection.** `checkUpkeep` does an O(circles × members) scan via `getEligibleAutomatedClaims()`. Add Chainlink's `cannotExecute` modifier (reverts when `tx.origin != address(0)`) so the function can only be called via off-chain `eth_call`:

```suggestion
  function checkUpkeep(bytes calldata)
    external
    view
    override
    cannotExecute
    returns (bool upkeepNeeded, bytes memory performData)
  {
    return _claimChecker();
  }
```

Requires inheriting `AutomationCompatible` (which also addresses Finding 2):

```solidity
import {AutomationCompatible} from "@chainlink/contracts/src/v0.8/automation/AutomationCompatible.sol";

contract AutomaticSavingCircles is IAutomaticSavingCircles, AutomationCompatible, Ownable, ReentrancyGuard { ... }
```

If you'd prefer to avoid the Chainlink dependency, drop a NatSpec line stating `checkUpkeep` is intended for off-chain simulation only.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -110,14 +110,25 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
...
-      catch (bytes memory reason) {
-        emit AutomatedClaimFailed(_circleIds[i], _members[i], reason);
-      }
+  /// @inheritdoc IAutomaticSavingCircles
+  function performUpkeep(bytes calldata _performData) external override nonReentrant onlyAutomationExecutor {
+    if (
+      _performData.length < 4
+        || bytes4(_performData[0:4]) != IAutomaticSavingCircles.batchExecuteAutomatedClaims.selector
+    ) {
+      revert InvalidPerformData();
```
franrolotti, line 122: **Finding 4 — defensive length check.** The `_performData.length < 4` branch is currently only exercised indirectly via the wrong-selector test. Worth a direct test (see suggestion on the test file) so a future refactor of the slicing logic can't silently regress this.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -141,11 +152,7 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
 
   /// @inheritdoc IAutomaticSavingCircles
   function claimChecker() external view override returns (bool canExec, bytes memory execPayload) {
-    (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedClaims();
-    canExec = _automationCanExecute(circleIds);
-    if (!canExec) return (false, bytes(''));
-
-    execPayload = abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedClaims, (circleIds, members));
+    return _claimChecker();
   }
```
franrolotti, line 157: **Finding 6 — naming.** Consider a one-liner here noting that `claimChecker()` is functionally equivalent to `checkUpkeep(bytes(''))` and retained for symmetry with `depositChecker()` (which is **not** wired into Chainlink — deposits still use the legacy resolver flow). Helps future readers who'll wonder why both exist.

src/contracts/AutomaticSavingCircles.sol
```diff
@@ -256,6 +270,28 @@ contract AutomaticSavingCircles is IAutomaticSavingCircles, Ownable, ReentrancyG
     SAVING_CIRCLES.withdrawFor(_circleId, _member);
   }
 
+  /**
+   * @dev Returns whether automatic claims can execute and the calldata Chainlink should perform
+   * @return canExec Whether automation can execute the claims
+   * @return execPayload Encoded calldata for the automated claim execution
+   */
+  function _claimChecker() internal view returns (bool canExec, bytes memory execPayload) {
+    (uint256[] memory circleIds, address[] memory members) = getEligibleAutomatedClaims();
```
franrolotti, line 279: **Finding 3 — `performGas` cliff.** `getEligibleAutomatedClaims` returns *all* eligible entries unbounded. Chainlink registrations cap `performGas` (default 500k, max 5M); once the eligible set outgrows that budget every `performUpkeep` OOGs and no claims land. Consider capping the returned arrays to a sensible N (or making N configurable) so each upkeep makes forward progress. Not introduced here, but the Chainlink move makes it more salient than under Gelato.

test/unit/AutomaticSavingCircles.t.sol
```diff
@@ -472,6 +472,67 @@ contract AutomaticSavingCirclesUnit is SavingCirclesTestBase {
...
+    uint256[] memory circleIds = new uint256[](0);
+    address[] memory targetMembers = new address[](0);
+    bytes memory performData =
+      abi.encodeCall(IAutomaticSavingCircles.batchExecuteAutomatedDeposits, (circleIds, targetMembers));
+
+    vm.prank(automationExecutor);
+    vm.expectRevert(abi.encodeWithSelector(IAutomaticSavingCircles.InvalidPerformData.selector));
+    automaticSavingCircles.performUpkeep(performData);
+  }
+
```
franrolotti, line 535: **Finding 4 — add a direct length-guard test.** Suggested addition alongside `test_PerformUpkeep_WhenPerformDataIsInvalid`:

```solidity
function test_PerformUpkeep_WhenPerformDataIsTooShort() external {
  vm.prank(automationExecutor);
  vm.expectRevert(abi.encodeWithSelector(IAutomaticSavingCircles.InvalidPerformData.selector));
  automaticSavingCircles.performUpkeep(bytes(''));

  vm.prank(automationExecutor);
  vm.expectRevert(abi.encodeWithSelector(IAutomaticSavingCircles.InvalidPerformData.selector));
  automaticSavingCircles.performUpkeep(hex"001122");
}
```

Also consider a stale-payload test: capture `performData`, advance time / have someone else claim, then call `performUpkeep` and assert `AutomatedClaimFailed` is emitted instead of a revert.

franrolotti (commented):
## Summary

Clean, well-scoped refactor. Mostly a naming/docs sweep (Gelato → automation/Chainlink) plus two new entry points (`checkUpkeep` / `performUpkeep`) that wrap existing eligibility + batch-claim logic. Solid coverage for the happy path, auth, and the invalid-payload case. Recommend approval with a few minor follow-ups below.

> _AI-generated review — please verify suggestions before applying._

## Strengths

- **Internal helper extraction.** Pulling `_batchExecuteAutomatedClaims` and `_claimChecker` out lets `performUpkeep`/`claimChecker` and `batchExecuteAutomatedClaims`/`checkUpkeep` share one implementation — no drift risk between the two entry points.
- **`performUpkeep` is permissioned** via `onlyAutomationExecutor` + `nonReentrant`. Matches Chainlink's recommended hardening of pinning `msg.sender` to the registered Forwarder so callers can't push arbitrary stale `performData`.
- **Stale-payload safety.** The `try/catch` around individual claims means a stale `performData` (already-claimed member, round advanced) emits `AutomatedClaimFailed` instead of reverting the batch — important since Chainlink can re-submit stale data between `checkUpkeep` and `performUpkeep`.
- **Selector check on `performData`** defends against operator misconfiguration (e.g. paste of a `batchExecuteAutomatedDeposits` payload), and there's a dedicated test for it.

## Findings

### 1. `checkUpkeep` not protected against on-chain calls — minor

Chainlink's `AutomationCompatible` exposes a `cannotExecute` modifier that reverts when `tx.origin != address(0)`, ensuring the function is only invoked off-chain via `eth_call`. `getEligibleAutomatedClaims()` iterates over all circles × members, so on-chain callers (griefers or honest mistakes) hitting `checkUpkeep` could waste a lot of gas in a single tx. See inline suggestion on `checkUpkeep`.

If you'd rather not pull the Chainlink dep in, the NatSpec should at least note that `checkUpkeep` is intended for off-chain simulation only.

### 2. No formal `AutomationCompatibleInterface` inheritance — nit

The contract implements the expected ABI by shape but doesn't inherit `AutomationCompatibleInterface`. Inheriting it gives compile-time signature safety and makes the integration self-documenting. Purely hygiene.

### 3. `performUpkeep` gas-bound by stale `circleIds.length` — consider a soft cap

Chainlink registrations have a `performGas` limit (default 500k, max 5M). `getEligibleAutomatedClaims` returns *all* eligible entries with no slicing. Once eligible-claim count grows past the per-tx gas budget, every `performUpkeep` will OOG and **no** claims will land — Chainlink will not auto-paginate. Two cheap options:

- Cap the returned arrays in `getEligibleAutomatedClaims` (e.g. to N items per call) so each upkeep makes progress.
- Or document that operators must size `performGas` to N × per-claim cost and accept the eventual scaling cliff.

Not introduced by this PR, but the move to Chainlink (which is stricter about `performGas`) makes it more salient.

### 4. Missing test: `performUpkeep` with `performData.length < 4` — minor

The length guard is tested only implicitly via the wrong-selector test. A direct test with `bytes('')` or `hex"00"` would lock in the length branch. See inline suggestion in the test file.

### 5. `automationExecutor` semantics changed without a setter migration — verify intent

In Chainlink Automation v2, the `msg.sender` of `performUpkeep` is the **Forwarder address**, derived per-upkeep registration and only known *after* registering. Deploy flow becomes:

1. Deploy `AutomaticSavingCircles` (no executor / placeholder).
2. Register upkeep on Chainlink → obtain forwarder.
3. Owner calls `setAutomationExecutor(forwarder)`.

Fine — rotation is already supported — but worth calling out in deploy docs since this differs from Gelato's "dedicated sender" pattern.

### 6. Naming nit — `claimChecker` / `depositChecker` retained

These are now redundant with `checkUpkeep` on the claim side. Keeping `claimChecker` is fine for symmetry with `depositChecker` (still resolver-style), but a one-liner noting that `claimChecker` ≡ `checkUpkeep` would help future readers. The PR description should also note that `depositChecker` is **not** wired into Chainlink — deposits still use the legacy resolver flow.

## Verdict

**Approve with minor follow-ups.** The refactor is mechanical, new entry points are small, and the safety properties (auth, reentrancy, per-target try/catch, selector validation) are all in place. Findings 1 and 3 are the only ones worth addressing before mainnet — both can be follow-ups if Sepolia behavior is already validated.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Applied the docs-only fallback in 811d774 — NatSpec on `checkUpkeep` now notes it is intended for off-chain simulation only. Pulling in the Chainlink dep for the formal `cannotExecute` modifier is out of scope for this PR; the on-chain risk is gas-griefing a view function, not funds.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Direct length-guard test added in e636e0c.

src/contracts/AutomaticSavingCircles.sol
franrolotti: NatSpec clarification added in ba729a9 — `claimChecker` is now documented as equivalent to `checkUpkeep`, and `depositChecker` is flagged as the legacy resolver flow not wired into Chainlink.

src/contracts/AutomaticSavingCircles.sol
franrolotti: Not introduced by this PR — the unbounded iteration in `getEligibleAutomatedClaims` predates the Chainlink migration. Worth tracking separately; opened #180.

test/unit/AutomaticSavingCircles.t.sol
franrolotti: Added in e636e0c (covers both `bytes('')` and `hex'00'` against the length guard).

franrolotti:
@RonTuretzky can I merge this?

PR #184, opened 2026-07-07
feat: add ASCA, Goal Savings, and Collective Fund stack types

RonTuretzky:
## Summary

This PR adds **three new stack types** alongside the existing ROSCA (`SavingCircles`), completing the family of community savings primitives:

| Contract | Stack type | One-liner |
|---|---|---|
| `AccumulatingSavingCircles` | **ASCA** (Accumulating Savings & Credit Association) | Members deposit *whatever amount, whenever*; each earns a credit line as a percentage of their own savings, borrows from the pool, and repays within a configured number of periods with optional interest — interest streams pro-rata to savers. |
| `GoalSavingCircles` | **Goal savings** | A group saves toward a target by a deadline. Funds are locked while funding (the commitment device); on success the pot releases to a pre-agreed beneficiary — or, with no beneficiary, everyone reclaims exactly their own contributions. On failure/cancel: full refunds. |
| `CollectiveFundCircles` | **Collective fund** (white-label communal pot) | A community saves together and *spends together*: flexible deposits, rage-quit anytime for a pro-rata share, one-member-one-vote disbursement proposals with a snapshotted electorate, permissionless execution. Fund name lives on-chain so any frontend can white-label a fund from just `?f=<id>`. |

Together with #183 (the chit fund / reverse-auction ROSCA), this covers ROSCA + Chit + ASCA + Goal + Collective.

## Design

All three follow the house pattern established by `SavingCircles`:

- Singleton registry keyed by `uint256 id`, upgradeable (`OwnableUpgradeable` + `ReentrancyGuardUpgradeable` + `EIP712Upgradeable`), `_disableInitializers()` constructor, `initialize(address owner)` behind a `TransparentUpgradeableProxy`.
- **Same EIP-712 `StacksInvite` domain** and `Invite(uint256 id,uint256 nonce)` typehash as `SavingCircles` — the app-stacks invite flow ports over unchanged (only `verifyingContract` differs).
- Admin ERC20 allowlist (`setTokenAllowed`), SafeERC20, custom errors only, checks-effects-interactions, `nonReentrant` on every token-moving external, interface-first with full NatSpec and `@inheritdoc`.
- Internal per-fund cash ledgers (never `balanceOf` for logic) — funds in the singleton cannot bleed into each other.

### Safety highlights

- **ASCA:** credit lines are capped at 100% of the borrower's own savings (`borrowLimitBps <= 10_000`), so every loan is fully self-collateralized — a defaulter can only ever burn their own savings, never another member's principal. Interest accrual is simple, per whole period, capped at the due date; overdue loans are **permissionlessly liquidatable** (pure ledger reallocation, no token movement). Interest distribution uses a MasterChef-style per-share accumulator (floor rounding, dust stays in the pool).
- **Goal:** no division anywhere — all accounting is exact addition/subtraction, so conservation is an equality. `goalReached`/`cancelled`/`released` are one-way latches; release is permissionless once funded.
- **Collective:** share conservation (`Σ sharesOf == totalShares`), pro-rata rage-quit maths round down in favor of the pool, the last exiter sweeps dust exactly. Proposals snapshot the electorate via an append-only member list (`memberIndex < electorate`, O(1) per proposal), require `ceil(electorate × thresholdBps / 10_000)` yes-votes, auto-pass early, and expire after one extra voting period. **Rage-quit is never blocked by any proposal state** — a member's deposit is never hostage to a vote they lost.

Design notes for each contract are in `docs/` (mechanism, invariants, judged decisions), mirroring the pattern from #183.

## Testing

```
forge test    # 437 passed (168 pre-existing + 269 new), 0 failed
```

- `AccumulatingSavingCirclesUnit` — 88 tests, 100% branch coverage (40/40)
- `GoalSavingCirclesUnit` — 80 tests, 100% branch coverage (20/20)
- `CollectiveFundCirclesUnit` — 101 tests, 100% branch coverage (37/37)

Each suite covers every revert branch, every state transition, event emissions, boundary timestamps, rounding/dust behavior, and cross-cutting conservation checks (including a two-funds-one-token no-bleed test). `forge fmt` + `solhint` clean.

The implementation went through an adversarial review pass (independent exploit-hunting and spec-conformance lenses per contract, findings PoC-verified against the code): **no confirmed findings**.

## Deployment / CI wiring

- `script/Common.sol` `_deployContracts` now deploys all three behind their own transparent proxies (existing return value unchanged, so `IntegrationBase` keeps working).
- `.github/upgrades.json`: three new entries for upgrade-safety validation.

## Scope (v1) & follow-ups

Ships the core mechanics above. Documented extensions (in `docs/`): loan top-ups & third-party repay + Chainlink liquidation satellite (ASCA), open-join & yield-bearing goals (Goal), member removal, weighted voting, earmarked donations, ERC-7572 `contractURI` metadata (Collective), viewer-lens getters and fuzz/invariant suites as fast-follows.

Frontend support for all three types is in the companion PR on `app-stacks`.

### DIFF
--- a/.github/upgrades.json
+++ b/.github/upgrades.json
@@ -2,7 +2,15 @@
   "contracts": [
     {
       "contract": "src/contracts/SavingCircles.sol:SavingCircles"
+    },
+    {
+      "contract": "src/contracts/AccumulatingSavingCircles.sol:AccumulatingSavingCircles"
+    },
+    {
+      "contract": "src/contracts/GoalSavingCircles.sol:GoalSavingCircles"
+    },
+    {
+      "contract": "src/contracts/CollectiveFundCircles.sol:CollectiveFundCircles"
     }
   ]
 }
-
--- a/docs/accumulating-saving-circles.md
+++ b/docs/accumulating-saving-circles.md
@@ -1,0 +1,102 @@
+# Accumulating Saving Circles (ASCA)
+
+*The "loose" cousin of the ROSCA: flexible savings, a self-collateralized credit line, and
+interest that flows back to the savers — the group banks itself.*
+
+This document is the design rationale for `src/contracts/AccumulatingSavingCircles.sol`. The
+definitive spec (storage layout, exact math, full test plan) lives in the design notes; the
+interface `src/interfaces/IAccumulatingSavingCircles.sol` is the API reference.
+
+---
+
+## 0. What problem it solves
+
+A `SavingCircles` ROSCA forces a fixed deposit on a fixed schedule and pays each member the pot
+exactly once. Many real groups (village associations, co-ops, friend circles) want the *other*
+classic mutual-finance shape instead — the **ASCA** (Accumulating Savings and Credit Association):
+each member puts in **whatever amount they want, whenever they want**, and in exchange earns a
+**credit line equal to a percentage of their own savings**. Loans are repaid within a configured
+number of periods, optionally with simple interest set by the group, and **all interest flows back
+pro-rata to the savers** rather than to a bank.
+
+## 1. The mechanism
+
+- **Singleton registry.** One contract hosts many funds, keyed by `uint256 id`, exactly like
+  `SavingCircles`. The organizer calls `create(token, borrowLimitBps, interestRateBps,
+  repaymentPeriods, periodLength)`, becomes the fund owner and its first member, and invites
+  members with the same EIP-712 `StacksInvite` flow the app already ships (domain
+  `'StacksInvite'/'1'`, typehash `Invite(uint256 id,uint256 nonce)`; only the verifying-contract
+  address differs).
+- **Savings.** `deposit`/`depositFor` add any positive amount to `savings[id][member]`; `withdraw`
+  returns it, subject to the collateral and liquidity guards below. Savings double as
+  MasterChef-style accumulator shares.
+- **Credit.** `borrow(id, amount)` requires `amount * 10_000 <= borrowLimitBps * savings` (the
+  multiplicative form leaves no rounding loophole) and enough `poolCash`. One loan at a time:
+  a second `borrow` reverts with `OutstandingLoan` until the first is fully repaid or liquidated.
+- **Interest.** Simple interest accrues lazily per **whole elapsed period** on outstanding
+  principal (`principal * interestRateBps / 10_000` per period), checkpointed at every touch, and
+  **capped at `repaymentPeriods` periods** — accrual stops at the due date, so the borrower's
+  liability is known at borrow time and can never outgrow deterministic bounds.
+- **Repay waterfall.** `repay` settles accrued interest first, then principal, pulls at most the
+  total debt (overpayment is not taken), and immediately streams the interest received to savers
+  via `accInterestPerShare += interest * 1e18 / totalSavings` (floor; sub-wei dust stays in the
+  pool). Members collect with `claimInterest`.
+- **Liquidation.** Once `block.timestamp >= startTime + repaymentPeriods * periodLength`, anyone
+  may call `liquidate`: it seizes the borrower's savings for the full principal (always covered),
+  plus `min(interestOwed, remaining savings)` which is distributed to savers as interest. It is a
+  **pure ledger reallocation** — no tokens move.
+- **Wind-down.** The organizer's one-way `deactivate` freezes `deposit`/`depositFor`/`borrow`/
+  `redeemInvite` while leaving `withdraw`/`repay`/`claimInterest`/`liquidate` open, so a
+  deactivated fund drains naturally. There is no share-out event; membership is continuous.
+
+## 2. Safety invariants
+
+Let, per fund: `S = totalSavings`, `B = totalBorrowed`, `C = poolCash`, `P = Σ pendingInterestOf`.
+
+- **I1 — internal cash ledger.** All logic reads `poolCash`, never `token.balanceOf`, so donations
+  and other funds sharing the contract cannot corrupt accounting. Every outbound transfer is
+  guarded by `poolCash >= amount`.
+- **I2 — full self-collateralization.** `borrowLimitBps <= 10_000` implies `principal <= savings`
+  for every open loan. Established at `borrow`, preserved by the `withdraw` guard
+  (`(principal + accrued interest) * 10_000 <= borrowLimitBps * remaining savings` — accrued
+  interest counts as debt), and cleared by `liquidate`. Liquidation can therefore always seize the
+  full principal: **a defaulter can only ever burn their own savings, never another member's
+  principal.**
+- **I3 — solvency.** `C + B >= S + P` holds inductively across every operation (distribution and
+  settlement round down, so rounding always favors the pool). Combined with I2, every receivable
+  in `B` is backed by seizable savings, so all liabilities are eventually payable and the
+  liquidity guards are belt-and-suspenders: honest members can always exit within at most
+  `repaymentPeriods * periodLength` via permissionless liquidation.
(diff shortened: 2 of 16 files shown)


franrolotti (changes requested):
## AI-assisted economic & security review

This review was generated with AI assistance (Claude Code) and reviewed by @franrolotti before posting. Methodology: full source read of all three new contracts, interfaces, and design docs; three independent deep audits (one per contract); direct verification of the highest-severity claims against the deployed source, including a working PoC for the critical finding. A fuller written version of this analysis (with worked numeric proofs for each invariant claim) exists as a companion document.

**Verdict: requesting changes.** One critical, PoC-confirmed fund-drain bug must be fixed before merge. Two further high-severity issues should be fixed or explicitly, consciously accepted. Architecture, invariant discipline (internal cash ledgers instead of `balanceOf`, consistent checks-effects-interactions, custom errors, shared EIP-712 invite flow across all four stack types), and test coverage (100% branch coverage on all three new contracts, confirmed via the CI coverage report) are genuinely strong — these are specific, fixable gaps, not a signal to redesign.

The PR description's claim of an "adversarial review pass ... no confirmed findings" does not hold up against the findings below.

---

### 🔴 Critical — `CollectiveFundCircles`: zero-share window lets any member steal a donation outright

`donate()` (`CollectiveFundCircles.sol:182-191`) adds to `poolBalance` without minting shares. `deposit()` (`:165-179`) mints shares 1:1 regardless of the current pool/share ratio. `withdraw()` (`:194-210`) pays `shares × poolBalance / totalShares`. Whenever a fund's `totalShares == 0` — which is exactly the state right after `create()`, before anyone has deposited, or any time after a fund's membership fully rage-quits — this breaks down.

**Exploit (PoC-confirmed with a passing `forge test`):** a donor sends 10,000 tokens into a fresh fund via `donate()` (poolBalance = 10,000, totalShares = 0). Any existing member calls `deposit(1 wei)` — they now hold 100% of totalShares. They immediately call `withdraw(1)`: payout = `1 × 10,001 / 1 = 10,001`. The member walks away with the entire donated pool for one wei of their own money.

This is realistic, not contrived — the design doc's own pitch is "seed a fund with a grant before recruiting members," and `donate()` is deliberately open to non-members for exactly this workflow. The test suite even constructs the precursor state (`test_PreviewWithdrawReturnsZeroWhenTotalSharesZero`) but never follows through with a deposit+withdraw, so existing tests walk right past it.

**Suggested fix:** require `totalShares > 0` as a precondition for `donate()`, mint a small permanently-locked "dead" share at `create()` (standard ERC-4626-style mitigation), or price the first deposit off `poolBalance` instead of a flat 1:1 mint.

---

### 🟠 High — `CollectiveFundCircles`: approval threshold has no floor — a fund creator's own vote can always be enough

`approvalThresholdBps` (`:120`) is validated only as `0 < x ≤ 10,000` at `create()`, then fixed forever. `requiredYes` (`:233`) is `ceil(electorate × thresholdBps / 10,000)`, and the proposer auto-votes yes (`:242`) the moment they call `propose()`. For any fund where `electorate ≤ 10,000 / thresholdBps`, `requiredYes` collapses to exactly 1 — at `thresholdBps = 1` (a legal value), that covers funds with up to 10,000 members.

`create()` is permissionless and the creator auto-joins as the first member. An attacker can create a fund with `thresholdBps = 1`, invite genuine members who deposit/donate trusting "one-member-one-vote" governance, then `propose()` sending the whole pool to themselves — their own auto-yes vote already satisfies quorum, with zero votes from anyone else and zero capital from the attacker. This is worth noting alongside the fact that shares (economic stake) and votes (governance) are completely decoupled in this contract by design (`sharesOf`/`totalShares` never appear in `propose`/`vote`/`execute`) — so there's no secondary stake-based check that would catch this.

**Suggested fix:** clamp `requiredYes = max(requiredYes, min(electorate, 2))` so that once a fund has more than one member, at least two genuine yes-votes are always required regardless of `thresholdBps`. This closes the single-vote path without reintroducing share-weighted voting (which would bring back the plutocracy problem one-member-one-vote was designed to avoid).

*(Lower severity than the item above because `getFund(id).approvalThresholdBps` is a public view — a careful depositor can inspect it before joining.)*

---

### 🟠 High — `AccumulatingSavingCircles`: interest-distribution sandwich

`accInterestPerShare` is updated atomically inside `repay()`, pro-rata to whatever `totalSavings` is at that exact block — no vesting, no time-weighting, no deposit/withdraw cooldown. A member watching for an imminent `repay()` can front-run it with a large `deposit()` and back-run it with `withdraw()` the moment interest is credited, capturing most of the interest payment despite zero real time-in-pool. Example: Alice is the sole saver for a week while a loan accrues 50 tokens of interest; Mallory deposits 100,000 tokens one block before `repay()` and withdraws right after, capturing ~98% of the interest that Alice's capital actually backed. This directly contradicts the design doc's "interest flows back pro-rata to the savers" claim and isn't covered by any existing test.

**Suggested fix:** a short withdrawal cooldown after deposit, or linear/streamed vesting of interest instead of instant crediting.

*(Bounded to invited members of a closed circle, not open MEV, which lowers but doesn't remove the risk.)*

---

### 🟡 Medium

- **`CollectiveFundCircles` — no timelock between quorum and execution.** Once `yesVotes ≥ requiredYes`, `execute()` is callable in the same transaction as the last yes-vote. The doc's minority-protection narrative ("rage-quit is never blocked by any proposal state") is code-level true but practically hollow if there's no on-chain window to react before a disbursement fires. Worth a conscious, documented decision either way.
- **`GoalSavingCircles` — a non-receiving beneficiary permanently locks the pot.** `release()` is the only exit path once `goalReached` latches with a non-zero beneficiary; there's no rescue path if the beneficiary is blacklisted by the token, is a non-receiving contract, or reverts. This is explicitly named as a deferred v1 trade-off in the design doc, so it's a known, accepted gap rather than an oversight — but deserves explicit sign-off given the blast radius, and frontends should vet beneficiary addresses before a group commits to one.

### ⚪ Low / nits

- `CollectiveFundCircles`: donated dust can be permanently stranded if `totalShares` hits 0 and nobody redeposits (same root cause as the critical finding, minus an attacker).
- `CollectiveFundCircles`: `withdraw()` never clears `isMember`, so a member who rage-quits to zero shares keeps permanent propose/vote rights with no remaining stake.
- `GoalSavingCircles`: no validation on the beneficiary address at `create()` (footgun, not exploit on its own).
- Two nits already flagged by the Copilot review (`getProposal`/`proposalState` revert `ProposalNotFound` instead of `FundNotFound` for a nonexistent fund ID) — cosmetic, worth picking up while in the file.

### What's solid (verified directly, not just claimed)

- ASCA's self-collateralization invariant (`principal ≤ savings` per open loan) holds under liquidation — worked through the ledger math and confirmed a defaulting borrower can only ever burn their own savings, never another member's principal.
- ASCA's per-member borrow cap, summed across all members, provides a full liquidity guarantee for non-borrowers: since `borrowLimitBps` is hard-capped at 100% (`MAX_BPS = 10_000`, enforced at `create()`), `poolCash ≥ Σ(non-borrowers' savings) + pendingInterest` always holds — a member who never borrows can always withdraw their full balance on demand, no matter how much others have borrowed.
- `GoalSavingCircles` accounting is exact add/subtract as claimed (no division anywhere), checks-effects-interactions is followed throughout, and the one-way state latches genuinely never reset.
- `CollectiveFundCircles` governance is one-member-one-vote (not share-weighted), so there's no flash-deposit vote-buying issue, and no state-changing function loops over the member list, so there's no unbounded-gas DoS as membership grows.

---

*Repo-wide, not specific to this PR: Slither static analysis is disabled in CI (`run-slither: false`), so none of the findings above would have been caught automatically.*

Issue #189, opened 2026-08-17
Epic: Yield-bearing savings circles (sDAI yield via crowdstake.fun)


BreadrichEngels:
## Goal

Make Stacks savings circles earn yield while they run: route deposits into crowdstake.fun's **sDAI** yield engine and pay **time-weighted** yield to members. The user-facing token stays **BREAD** (wrapped to sDAI under the hood).

## Yield model (agreed)

Two periods per round — an **accrual window** and a **claim**:

- **Base = per-user time-weighted accrual (Model B):** each depositor accrues **their own** yield; **earlier deposit = more yield**.
- **Claimant option (Model A element):** each round's designated recipient takes the principal pot, and **may leave it staked to keep compounding** instead of claiming immediately (claim = funds out, yield stops on those funds).
- **Claim-timing is modular (circle-setup flag):** yield is claimable either **(A) only when the whole circle finishes**, or **(B) anytime**.
- **No forfeiture** — defaulters are not slashed (explicitly out of scope for v1).

### Flow (per round)
1. **Accrual window** — members deposit BREAD → wrapped BREAD → xDAI → **sDAI**, staked. Each deposit accrues **time-weighted yield to its depositor**.
2. **Round close → claim** — the round's recipient can **claim the pot** (funds out, stops earning) **or leave it staked** to keep compounding.
3. **Reset** — next round; members **re-deposit** to keep accruing.
4. **End** — everyone has taken the pot once; yield is claimable per the circle's claim-timing setting.

## Reuse from crowdstake.fun (already built + live on Gnosis)

- **sDAI adapter:** `SexyDaiYield` / `PoolNativeYield` (xDAI/WXDAI → sDAI ERC-4626), deployed (`contracts/deployments/gnosis.json`).
- **Time-weighting:** `TimeWeightedVotingPower` (OZ checkpoints).
- **Harvest/distribute:** `BaseDistributionManager.claimAndDistribute()` + strategies.
- **Interface:** `IYieldModule` (`mint` / `burn` / `claimYield` / `yieldAccrued`).

## Build (new work — the gaps)

1. **Yield-bearing custody in `SavingCircles.sol`** — on deposit, route funds into the sDAI pool (`IYieldModule.mint`); on payout, redeem (`burn`). (Today it holds idle ERC-20 balances.)
[...]

PR #190, opened 2026-08-17
feat(yield): draft yield-bearing saving circles via crowdstake sDAI

BreadrichEngels:
Draft implementation for #189 — routing saving-circle deposits into crowdstake.fun's sDAI yield module.

> ⚠️ **Written without a local Solidity toolchain** (the agent sandbox ran out of disk, so Foundry/deps wouldn't install). This has **not been compiled or tested locally** — this PR is opened as a draft specifically so **CI compiles it** and surfaces issues. Expect a round-trip or two. Do not merge without a build, tests, and an accounting/security review.

## What's here
- **`src/interfaces/IYieldModule.sol`** — crowdstake.fun yield-module interface (`mint` / `burn` / `claimYield` / `yieldAccrued`).
- **`src/contracts/SavingCircles.sol`** — marked `_deposit` / `_withdraw` / `decommission` as `virtual` so the yield variant can hook them. Storage layout unchanged.
- **`src/contracts/YieldSavingCircles.sol`** — the yield layer (inherits `SavingCircles`).

## Model implemented (per #189)
- **Per-member, time-weighted accrual** via a MasterChef-style global accumulator over staked principal → earlier deposits earn more.
- **Claimant option**: `claim(id, keepStaked)` — take the pot to your wallet, or leave it staked so it keeps compounding for you (`withdrawStakedPrincipal` later).
- **Modular claim-timing** per circle (`ClaimTiming.Anytime` vs `EndOnly`) via `setClaimTiming`.
- **No forfeiture** (out of scope for v1).

## Known open items (need a compiler + review)
- **Shared-pool attribution.** Accounting is contract-global (one shared pool). Per-circle isolation and whether to use the pool's native per-holder split (`keptYieldOf`) instead of a local accumulator is an open design call.
- **BREAD wrap.** v1 treats the circle token as the module's underlying (identity). Keeping BREAD user-facing (BREAD→xDAI→sDAI and back) is stubbed behind `_toUnderlying` / `_fromUnderlying` seams and still needs implementing.
- **Upgrade path / `AutomaticSavingCircles` lockstep / decommission yield-unwind** — see #189 gaps.
- **No tests yet** — the mock-yield-module test suite is pending toolchain/CI.

Closes none yet; tracks #189.

### DIFF
--- a/src/contracts/SavingCircles.sol
+++ b/src/contracts/SavingCircles.sol
@@ -153,7 +153,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   }
 
   /// @inheritdoc ISavingCircles
-  function decommission(uint256 _id) external override nonReentrant onlyActive(_id) onlyMember(_id, msg.sender) {
+  function decommission(uint256 _id) external virtual override nonReentrant onlyActive(_id) onlyMember(_id, msg.sender) {
     if (!_isDecommissionable(_id)) revert NotDecommissionable();
 
     address token = circles[_id].token;
@@ -379,7 +379,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
    * @dev Make a withdrawal from a specified circle.
    *      Anyone can trigger the payout for a valid member whose turn is withdrawable.
    */
-  function _withdraw(uint256 _id, address _member) internal onlyMember(_id, _member) {
+  function _withdraw(uint256 _id, address _member) internal virtual onlyMember(_id, _member) {
     Circle storage _circle = circles[_id];
 
     if (!_claimable(_id, _member)) revert NotWithdrawable();
@@ -406,7 +406,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
     uint256 _id,
     uint256 _value,
     address _member
-  ) internal onlyCommissioned(_id) onlyMember(_id, _member) {
+  ) internal virtual onlyCommissioned(_id) onlyMember(_id, _member) {
     Circle memory _circle = circles[_id];
 
     if (block.timestamp < circles[_id].effectiveCircleStartTime) {
--- a/src/contracts/YieldSavingCircles.sol
+++ b/src/contracts/YieldSavingCircles.sol
@@ -1,0 +1,277 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {IERC20} from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import {SafeERC20} from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+
+import {SavingCircles} from 'contracts/SavingCircles.sol';
+import {IYieldModule} from 'interfaces/IYieldModule.sol';
+
+using SafeERC20 for IERC20;
+
+/**
+ * @title YieldSavingCircles
+ * @notice A yield-bearing ROSCA. Deposits are routed into a crowdstake.fun yield
+ *         module (sDAI on Gnosis) so principal earns yield while the circle runs.
+ * @author Breadchain Collective
+ * @dev  ============================ DRAFT / NOT YET COMPILED ============================
+ *       Written without a local Solidity toolchain (agent sandbox was out of disk).
+ *       Compilation, tests, and an accounting/security review are REQUIRED before use.
+ *       Tracking issue: BreadchainCoop/saving-circles#189.
+ *       ================================================================================
+ *
+ *       Agreed yield model (issue #189):
+ *       - Base = per-member, time-weighted accrual. Each deposit earns yield for the
+ *         depositor for as long as it sits staked -> earlier deposits earn more. This is
+ *         implemented with a MasterChef-style global accumulator over staked principal.
+ *       - Claimant option: when a member claims their round's pot they may either take it
+ *         (redeem principal to their wallet, yield on it stops) or LEAVE IT STAKED so the
+ *         pot keeps compounding for THEM until they pull it (`claim(id, keepStaked=true)`).
+ *       - Claim-timing is modular per circle: yield claimable ANYTIME or only ONCE THE
+ *         CIRCLE ENDS ({ClaimTiming}). Set via {setClaimTiming} before the circle starts.
+ *       - No forfeiture (out of scope for v1).
+ *
+ *       Token model: v1 treats the circle token as the yield module's underlying
+ *       (identity wrap). To keep BREAD user-facing (burn BREAD -> xDAI -> stake to sDAI,
+ *       and reverse on payout) override {_toUnderlying}/{_fromUnderlying}; the accounting
+ *       below is denominated in the module's underlying units. See #189 gap "BREAD wrap".
+ */
+contract YieldSavingCircles is SavingCircles {
+  /// @notice When a member may claim their accrued yield.
+  enum ClaimTiming {
+    Anytime,
+    EndOnly
+  }
+
+  uint256 private constant _ACC_PRECISION = 1e18;
+
+  /// @notice The crowdstake.fun yield module (stake pool) principal is staked into.
+  IYieldModule public yieldModule;
+
+  /// @notice Per-circle yield claim-timing policy. Defaults to Anytime (0).
+  mapping(uint256 id => ClaimTiming timing) public claimTiming;
+
+  // --- MasterChef-style global yield accounting (single shared pool) ---
+  uint256 private _accYieldPerPrincipal; // scaled by _ACC_PRECISION
+  uint256 private _totalPrincipal; // total principal currently staked across all circles
+  uint256 private _lastYieldSeen; // module yieldAccrued() already booked into the accumulator
+  mapping(address member => uint256 principal) private _principalOf;
+  mapping(address member => uint256 debt) private _rewardDebt;
+  mapping(address member => uint256 amount) public claimableYield; // credited, awaiting claim
+
+  /// @dev Per-call flag: whether the claimant keeps their pot staked. Reset after use.
+  bool private _keepStakedFlag;
+
+  event YieldModuleSet(address indexed yieldModule);
+  event ClaimTimingSet(uint256 indexed id, ClaimTiming timing);
+  event PotKeptStaked(uint256 indexed id, address indexed member, uint256 amount);
+  event YieldClaimed(address indexed member, address indexed receiver, uint256 amount);
+
+  error YieldModuleNotSet();
+  error YieldLocked();
+  error NothingToClaim();
+
+  /// @notice Set the yield module. Owner only. Must be set before any deposits.
+  function setYieldModule(address _yieldModule) external onlyOwner {
+    if (_yieldModule == address(0)) revert YieldModuleNotSet();
+    yieldModule = IYieldModule(_yieldModule);
+    emit YieldModuleSet(_yieldModule);
+  }
+
+  /// @notice Set a circle's yield claim-timing. Owner of the circle, before it starts.
+  function setClaimTiming(uint256 _id, ClaimTiming _timing) external {
+    if (circles[_id].owner != msg.sender) revert NotOwner();
+    if (isActive[_id]) revert AlreadyActive();
+    claimTiming[_id] = _timing;
+    emit ClaimTimingSet(_id, _timing);
+  }
+
+  /**
+   * @notice Claim the round pot, choosing whether to keep it staked and compounding.
+   * @param _id The circle id.
+   * @param _keepStaked If true, the pot stays in the yield module earning for the caller
+   *        (withdraw later via {withdrawStakedPrincipal}); if false it is redeemed to the wallet.
+   */
+  function claim(uint256 _id, bool _keepStaked) external nonReentrant onlyMember(_id, msg.sender) onlyActive(_id) {
+    _keepStakedFlag = _keepStaked;
+    _withdraw(_id, msg.sender);
+    _keepStakedFlag = false;
+  }
+
+  /// @notice Claim accrued yield to `_receiver`, subject to the member's circles' claim-timing.
+  function claimYield(address _receiver) external nonReentrant {
+    if (address(yieldModule) == address(0)) revert YieldModuleNotSet();
+    _harvestGlobal();
+    _settle(msg.sender);
+    if (!_yieldUnlocked(msg.sender)) revert YieldLocked();
+
+    uint256 _amount = claimableYield[msg.sender];
+    if (_amount == 0) revert NothingToClaim();
+    claimableYield[msg.sender] = 0;
+
+    // Pull yield (underlying) from the module, then hand it to the member.
+    yieldModule.claimYield(_amount, address(this));
+    _lastYieldSeen -= _amount; // module's yieldAccrued() dropped by _amount
+    uint256 _out = _fromUnderlying(_amount);
+    _payoutToken().safeTransfer(_receiver, _out);
+
+    emit YieldClaimed(msg.sender, _receiver, _out);
+  }
+
+  /// @notice Withdraw principal a claimant previously left staked (keepStaked path).
+  function withdrawStakedPrincipal(address _receiver) external nonReentrant {
+    if (address(yieldModule) == address(0)) revert YieldModuleNotSet();
+    _harvestGlobal();
+    _settle(msg.sender);
+
+    uint256 _principal = _principalOf[msg.sender];
+    if (_principal == 0) revert NothingToClaim();
+    _principalOf[msg.sender] = 0;
+    _totalPrincipal -= _principal;
+    _rewardDebt[msg.sender] = 0;
+
+    yieldModule.burn(_principal, address(this));
+    uint256 _out = _fromUnderlying(_principal);
+    _payoutToken().safeTransfer(_receiver, _out);
+  }
+
+  // =========================================================================
+  //                              Overrides
+  // =========================================================================
+
+  /// @dev Route the just-deposited principal into the yield module and book it for the member.
+  function _deposit(uint256 _id, uint256 _value, address _member) internal override {
+    if (address(yieldModule) == address(0)) revert YieldModuleNotSet();
+
+    _harvestGlobal();
+    _settle(_member);
+
+    // Base pulls `_value` of the circle token into this contract and updates ROSCA bookkeeping.
+    super._deposit(_id, _value, _member);
+
+    uint256 _underlyingAmount = _toUnderlying(_value);
+    _underlyingToken().forceApprove(address(yieldModule), _underlyingAmount);
+    yieldModule.mint(_underlyingAmount, address(this));
+
+    _principalOf[_member] += _underlyingAmount;
+    _totalPrincipal += _underlyingAmount;
+    _rewardDebt[_member] = (_principalOf[_member] * _accYieldPerPrincipal) / _ACC_PRECISION;
+  }
+
+  /**
+   * @dev Claimant takes their round's pot. Each depositor's yield on that round is settled to
+   *      them (kept per Model B); the pot principal either redeems to the claimant, or (Model A)
+   *      is re-attributed to the claimant and kept staked to keep compounding.
+   */
+  function _withdraw(uint256 _id, address _member) internal override onlyMember(_id, _member) {
+    if (!_claimable(_id, _member)) revert NotWithdrawable();
+
+    _harvestGlobal();
+
+    uint256 _round = _memberStates[_id][_member].memberIndex; // claimant's turn == their index
+    address[] memory _members = circleMembers[_id];
+    uint256 _pot;
+
+    for (uint256 i = 0; i < _members.length; i++) {
+      address _depositor = _members[i];
+      uint256 _d = roundDeposits[_id][_round][_depositor];
+      if (_d == 0) continue;
+
+      uint256 _underlyingD = _toUnderlying(_d);
+
+      // Credit this depositor the yield their principal earned, then remove this round's principal.
+      _settle(_depositor);
+      _principalOf[_depositor] -= _underlyingD;
+      _totalPrincipal -= _underlyingD;
+      _rewardDebt[_depositor] = (_principalOf[_depositor] * _accYieldPerPrincipal) / _ACC_PRECISION;
+
+      roundDeposits[_id][_round][_depositor] = 0; // consumed into the pot
+      _pot += _underlyingD;
+    }
+
+    _memberStates[_id][_member].hasClaimed = true;
+
+    if (_keepStakedFlag) {
+      // Pot never leaves the pool; re-attribute it to the claimant so it keeps earning for them.
+      _settle(_member);
+      _principalOf[_member] += _pot;
+      _totalPrincipal += _pot;
+      _rewardDebt[_member] = (_principalOf[_member] * _accYieldPerPrincipal) / _ACC_PRECISION;
+      emit PotKeptStaked(_id, _member, _pot);
+    } else {
+      yieldModule.burn(_pot, address(this));
+      uint256 _out = _fromUnderlying(_pot);
+      _payoutToken().safeTransfer(_member, _out);
+    }
+
+    emit FundsWithdrawn(_id, _member, _pot);
+
+    if (_allMembersClaimed(_id)) {
+      isActive[_id] = false;
+    }
+  }
+
+  // =========================================================================
+  //                              Internals
+  // =========================================================================
+
+  /// @dev Book newly accrued module yield into the global per-principal accumulator.
+  function _harvestGlobal() internal {
+    if (_totalPrincipal == 0) return;
+    uint256 _accruedNow = yieldModule.yieldAccrued();
+    if (_accruedNow > _lastYieldSeen) {
+      uint256 _delta = _accruedNow - _lastYieldSeen;
+      _accYieldPerPrincipal += (_delta * _ACC_PRECISION) / _totalPrincipal;
+      _lastYieldSeen = _accruedNow;
+    }
+  }
+
+  /// @dev Move a member's pending accumulator yield into their claimable balance.
+  function _settle(address _member) internal {
+    uint256 _accumulated = (_principalOf[_member] * _accYieldPerPrincipal) / _ACC_PRECISION;
+    uint256 _pending = _accumulated - _rewardDebt[_member];
+    if (_pending > 0) claimableYield[_member] += _pending;
+    _rewardDebt[_member] = _accumulated;
+  }
+
+  /// @dev A member's yield is locked while any of their circles is EndOnly and still active.
+  function _yieldUnlocked(address _member) internal view returns (bool) {
+    uint256[] memory _ids = memberCircles[_member];
+    for (uint256 i = 0; i < _ids.length; i++) {
+      if (isActive[_ids[i]] && claimTiming[_ids[i]] == ClaimTiming.EndOnly) return false;
+    }
+    return true;
+  }
+
+  // =========================================================================
+  //      Wrap seams — override to keep BREAD user-facing (BREAD<->xDAI<->sDAI)
+  // =========================================================================
+
+  /// @dev Convert a circle-token amount into the yield module's underlying. Identity in v1.
+  function _toUnderlying(uint256 _amount) internal virtual returns (uint256) {
+    return _amount;
+  }
+
+  /// @dev Convert an underlying amount back into the circle/payout token. Identity in v1.
+  function _fromUnderlying(uint256 _amount) internal virtual returns (uint256) {
+    return _amount;
+  }
+
+  /// @dev The ERC20 the yield module pulls on mint / returns on burn. v1: the payout token.
+  function _underlyingToken() internal view virtual returns (IERC20) {
+    return _payoutToken();
+  }
+
+  /// @dev The ERC20 members are paid in (principal + yield). v1: single supported token.
+  function _payoutToken() internal view virtual returns (IERC20) {
+    return IERC20(_soleToken);
+  }
+
+  /// @dev v1 single-token constraint: set on first allowed token. See #189 (multi-token later).
+  address private _soleToken;
+
+  /// @notice Owner sets the single supported ERC20 for circles + payouts (v1 constraint).
+  function setSoleToken(address _token) external onlyOwner {
+    _soleToken = _token;
+  }
+}
--- a/src/interfaces/IYieldModule.sol
+++ b/src/interfaces/IYieldModule.sol
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+/**
+ * @title IYieldModule
+ * @notice Minimal surface of a crowdstake.fun yield module / stake pool that the
+ *         saving-circles integration consumes.
+ * @dev Mirrors `BreadchainCoop/crowdstake.fun` `contracts/src/interfaces/IYieldModule.sol`.
+ *      The concrete Gnosis implementation is `PoolNativeYield` / `SexyDaiYield`
+ *      (WXDAI/xDAI -> sDAI ERC-4626), deployed per `contracts/deployments/gnosis.json`.
+ *
+ *      Deposit model: `mint(amount, receiver)` pulls `amount` of the module's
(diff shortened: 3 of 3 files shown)


ss251:
## Bug: `IYieldModule.mint(uint256,address)` is the wrong ABI

`mint(amount, receiver)` will not hit the live crowdstake sDAI module. The deployed contracts use `mint(receiver, amount)` — different selector, so every deposit reverts.

### Copied interface (stale)

```solidity
function mint(uint256 amount, address receiver) external; // 0x94bf804d
```

- This PR's [`IYieldModule`](https://github.com/BreadchainCoop/saving-circles/blob/feat/yield-bearing-sdai/src/interfaces/IYieldModule.sol#L20)
- Copied from [crowdstake `IYieldModule`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/interfaces/IYieldModule.sol#L12)

### What is actually deployed (`SexyDaiYield` / `PoolNativeYield`)

```solidity
function mint(address receiver, uint256 amount) external; // 0x40c10f19
```

- [`AbstractToken.mint`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/abstract/AbstractToken.sol#L197)
- [`AbstractStakePool.mint`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/implementation/pool/AbstractStakePool.sol#L194)
- [`IToken`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/interfaces/IToken.sol#L8) / [`IStakePool`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/interfaces/IStakePool.sol#L30)
- [frontend ABI](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/src/lib/abis/token.ts#L236-L243)

### The call in this PR

```solidity
yieldModule.mint(_underlyingAmount, address(this)); // mint(uint256,address)
```

[`YieldSavingCircles.sol#L154`](https://github.com/BreadchainCoop/saving-circles/blob/feat/yield-bearing-sdai/src/contracts/YieldSavingCircles.sol#L154)

No `mint(uint256,address)` exists on the sDAI module, so this reverts.

### Why the copied interface is misleading

Crowdstake's DistributionManager is the only consumer of `IYieldModule`, and it only calls `yieldAccrued()` and `claimYield(amount, receiver)` — [never `mint()`](https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/base/BaseDistributionManager.sol#L148-L152). `burn` / `claimYield` already share `(amount, receiver)` on both sides, so those calls here would ABI-match. Only `mint` is reversed (ERC-4626 / OZ `mint(to, amount)` order).

### Fix

Call `mint(address(this), _underlyingAmount)` against `IToken` / `IStakePool`, not `IYieldModule`.

src/interfaces/IYieldModule.sol
```diff
@@ -0,0 +1,30 @@
...
+ *
+ *      Deposit model: `mint(amount, receiver)` pulls `amount` of the module's
+ *      UNDERLYING asset from the caller and credits `receiver` with principal.
+ *      `burn(amount, receiver)` redeems `amount` of principal and sends the
+ *      underlying to `receiver`. Yield accrues on top of principal and is read
+ *      via `yieldAccrued()` and pulled via `claimYield(amount, receiver)`.
+ */
+interface IYieldModule {
+  /// @notice Stake `amount` of the underlying (pulled from caller) crediting `receiver`.
+  function mint(uint256 amount, address receiver) external;
```
ss251: **1. Interface drift.** This `mint(uint256,address)` selector (`0x94bf804d`) is not what the concrete CrowdStake token/pool implements.

Live surface is `mint(address receiver, uint256 amount)` (`0x40c10f19`):
- https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/abstract/AbstractToken.sol#L197
- https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/implementation/pool/AbstractStakePool.sol#L194
- https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/interfaces/IToken.sol#L8
- https://github.com/BreadchainCoop/crowdstake.fun/blob/main/contracts/src/interfaces/IStakePool.sol#L30

CrowdStake's own `IYieldModule` is stale and unused for deposits (`DistributionManager` never calls `mint`). Do not copy it as the integration ABI. Use `IToken` / `IStakePool`.

```suggestion
      Deposit model: `mint(receiver, amount)` pulls `amount` of the module's
      UNDERLYING asset from the caller and credits `receiver` with principal.
      `burn(amount, receiver)` redeems `amount` of principal and sends the
      underlying to `receiver`. Yield accrues on top of principal and is read
      via `yieldAccrued()` and pulled via `claimYield(amount, receiver)`.
 */
interface IYieldModule {
  /// @notice Stake `amount` of the underlying (pulled from caller) crediting `receiver`.
  function mint(address receiver, uint256 amount) external;
```

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+
+    _harvestGlobal();
+    _settle(_member);
+
+    // Base pulls `_value` of the circle token into this contract and updates ROSCA bookkeeping.
+    super._deposit(_id, _value, _member);
+
+    uint256 _underlyingAmount = _toUnderlying(_value);
+    _underlyingToken().forceApprove(address(yieldModule), _underlyingAmount);
+    yieldModule.mint(_underlyingAmount, address(this));
```
ss251: **1. Wrong `mint` ABI.** This encodes `mint(uint256,address)`. The live CrowdStake module has `mint(address,uint256)` only, so every deposit reverts.

Must ship together with the interface change above.

```suggestion
    yieldModule.mint(address(this), _underlyingAmount);
```

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+
+  /// @dev Book newly accrued module yield into the global per-principal accumulator.
+  function _harvestGlobal() internal {
+    if (_totalPrincipal == 0) return;
+    uint256 _accruedNow = yieldModule.yieldAccrued();
+    if (_accruedNow > _lastYieldSeen) {
+      uint256 _delta = _accruedNow - _lastYieldSeen;
+      _accYieldPerPrincipal += (_delta * _ACC_PRECISION) / _totalPrincipal;
+      _lastYieldSeen = _accruedNow;
+    }
```
ss251: **2. Wrong yield source.** `yieldAccrued()` is CrowdStake's *donated/distributable* pool, not this contract's holder yield.

CrowdStake splits:
- `yieldAccrued()` — donated, claimer-only
- `totalYieldAccrued()` — full vault surplus
- `keptYieldOf(account)` / `claimKeptYield()` — holder-owned
- `setYieldSplit()` — keep vs donate

If this module is shared, other holders' donated yield enters `_delta` while their principal is not in `_totalPrincipal`. SavingCircles needs an explicit ownership path (dedicated instance + claimer, or 100% kept-yield as the holder) before this accumulator is meaningful.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+
+    uint256 _amount = claimableYield[msg.sender];
+    if (_amount == 0) revert NothingToClaim();
+    claimableYield[msg.sender] = 0;
+
+    // Pull yield (underlying) from the module, then hand it to the member.
+    yieldModule.claimYield(_amount, address(this));
+    _lastYieldSeen -= _amount; // module's yieldAccrued() dropped by _amount
+    uint256 _out = _fromUnderlying(_amount);
+    _payoutToken().safeTransfer(_receiver, _out);
```
ss251: **3 + 4. Unauthorized claimer, and the wrong asset comes back even if it were authorized.**

`AbstractToken.claimYield` / pool `claimYield` revert unless `msg.sender == yieldClaimer`. Standard CrowdStake wiring sets `DistributionManager` as that claimer. This PR never makes `YieldSavingCircles` the claimer.

If it *did* succeed:
- token-mode `SexyDaiYield.claimYield` *mints the CrowdStake token* to the receiver
- pool-mode pays WXDAI

Neither is "underlying ERC-20 that `_fromUnderlying` + `_payoutToken().safeTransfer` can handle" in the general case. This cannot assume the donated-yield claim path.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+      // Pot never leaves the pool; re-attribute it to the claimant so it keeps earning for them.
+      _settle(_member);
+      _principalOf[_member] += _pot;
+      _totalPrincipal += _pot;
+      _rewardDebt[_member] = (_principalOf[_member] * _accYieldPerPrincipal) / _ACC_PRECISION;
+      emit PotKeptStaked(_id, _member, _pot);
+    } else {
+      yieldModule.burn(_pot, address(this));
+      uint256 _out = _fromUnderlying(_pot);
+      _payoutToken().safeTransfer(_member, _out);
```
ss251: **4. `burn` remits native xDAI, not an ERC-20.** Both `SexyDaiYield._remit` and `PoolNativeYield._remit` do `sDAI -> WXDAI -> native xDAI` and `_nativeTransfer(receiver_, amount_)`.

This contract has no `receive()` / `fallback()`, so the native transfer reverts. Even if it accepted ETH, `_fromUnderlying` + `_payoutToken().safeTransfer` would still treat a native credit as if an ERC-20 arrived.

Same bug on `withdrawStakedPrincipal` (`burn` at L133). Principal vs yield vs token-mode claim are three different asset flows; a single `_fromUnderlying(uint256)` seam cannot represent them.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+    uint256 _round = _memberStates[_id][_member].memberIndex; // claimant's turn == their index
+    address[] memory _members = circleMembers[_id];
+    uint256 _pot;
+
+    for (uint256 i = 0; i < _members.length; i++) {
+      address _depositor = _members[i];
+      uint256 _d = roundDeposits[_id][_round][_depositor];
+      if (_d == 0) continue;
+
+      uint256 _underlyingD = _toUnderlying(_d);
```
ss251: **5. `_toUnderlying` is not a safe BREAD-adapter hook.** It is invoked here on *historical* `roundDeposits` during withdraw, and again on a live deposit at L152.

BREAD conversion is stateful (`BREAD burn -> native xDAI`, reverse on payout). If an override burns BREAD, this path burns the same historical amount a second time.

Split:
- asset movement/conversion (deposit/payout only)
- amount/accounting conversion (pure, idempotent, safe to re-run on history)

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+
+      // Credit this depositor the yield their principal earned, then remove this round's principal.
+      _settle(_depositor);
+      _principalOf[_depositor] -= _underlyingD;
+      _totalPrincipal -= _underlyingD;
+      _rewardDebt[_depositor] = (_principalOf[_depositor] * _accYieldPerPrincipal) / _ACC_PRECISION;
+
+      roundDeposits[_id][_round][_depositor] = 0; // consumed into the pot
```
ss251: **6. Do not consume `roundDeposits`.** Inherited `_deposit` / `_isDecommissionable` treat these values as "was the previous round fully funded?".

Zeroing them after the round-0 claim makes round 1 look like an incomplete previous round → `CircleTimedOut()` / decommissionable. Keep `roundDeposits` as immutable ROSCA history and track yield-principal redemption in separate storage.

```suggestion
```

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+
+    uint256 _principal = _principalOf[msg.sender];
+    if (_principal == 0) revert NothingToClaim();
+    _principalOf[msg.sender] = 0;
+    _totalPrincipal -= _principal;
+    _rewardDebt[msg.sender] = 0;
+
+    yieldModule.burn(_principal, address(this));
+    uint256 _out = _fromUnderlying(_principal);
+    _payoutToken().safeTransfer(_receiver, _out);
```
ss251: **7. This withdraws *all* of the caller's principal, not the keep-staked pot.**

`_deposit` also increments `_principalOf[member]`. There is no `circleId` and no `keptPotPrincipal`. A member can pull deposits that are still backing active rounds; `roundDeposits` still says those contributions exist, then the later pot claim subtracts that principal again.

Needs per-circle balances: mandatory deposits vs voluntarily retained pot. `withdrawStakedPrincipal(id, receiver)` should burn only `keptPotPrincipal[id][msg.sender]`.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+   * @notice Claim the round pot, choosing whether to keep it staked and compounding.
+   * @param _id The circle id.
+   * @param _keepStaked If true, the pot stays in the yield module earning for the caller
+   *        (withdraw later via {withdrawStakedPrincipal}); if false it is redeemed to the wallet.
+   */
+  function claim(uint256 _id, bool _keepStaked) external nonReentrant onlyMember(_id, msg.sender) onlyActive(_id) {
+    _keepStakedFlag = _keepStaked;
+    _withdraw(_id, msg.sender);
+    _keepStakedFlag = false;
+  }
```
ss251: **8. Claimant does not control keep-staked.** `_keepStakedFlag` is only set here. `SavingCircles.withdrawFor(id, member)` (and `AutomaticSavingCircles`) still call `_withdraw` with the flag false, so anyone — including automation — can force the pot out of the yield position.

The keep-staked choice has to live in per-circle/per-member storage (or `withdrawFor` must be overridden / disabled), not a transient flag on this one entrypoint.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+  /// @notice Per-circle yield claim-timing policy. Defaults to Anytime (0).
+  mapping(uint256 id => ClaimTiming timing) public claimTiming;
+
+  // --- MasterChef-style global yield accounting (single shared pool) ---
+  uint256 private _accYieldPerPrincipal; // scaled by _ACC_PRECISION
+  uint256 private _totalPrincipal; // total principal currently staked across all circles
+  uint256 private _lastYieldSeen; // module yieldAccrued() already booked into the accumulator
+  mapping(address member => uint256 principal) private _principalOf;
+  mapping(address member => uint256 debt) private _rewardDebt;
+  mapping(address member => uint256 amount) public claimableYield; // credited, awaiting claim
```
ss251: **9 + 12. Member state cannot be global if claim policy is per circle; and this mapping is not a UI-safe read.**

A shared `_accYieldPerPrincipal` can be fine if every circle shares one yield source. `_principalOf` / `_rewardDebt` / `claimableYield` cannot: a member in an `Anytime` circle and an `EndOnly` circle gets one collapsed `claimableYield[alice]`. `_yieldUnlocked` then blocks *all* of their yield if any active circle is `EndOnly`.

Need at least:
```solidity
principal[circleId][member]
rewardDebt[circleId][member]
claimableYield[circleId][member]
keptPotPrincipal[circleId][member]
```

`claimableYield` is also *settled storage*. Fresh sDAI yield only appears after `_harvestGlobal` + `_settle`. The UI reading this mapping and treating `0` as "nothing to claim" will disable claim while unsettled yield exists. Expose a view `pendingYield(circleId, member)` that simulates the accumulator.

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  /// @dev A member's yield is locked while any of their circles is EndOnly and still active.
+  function _yieldUnlocked(address _member) internal view returns (bool) {
+    uint256[] memory _ids = memberCircles[_member];
+    for (uint256 i = 0; i < _ids.length; i++) {
+      if (isActive[_ids[i]] && claimTiming[_ids[i]] == ClaimTiming.EndOnly) return false;
+    }
+    return true;
+  }
```
ss251: **9. `ClaimTiming` is broken by global member state.** Any active `EndOnly` circle locks the member's entire `claimableYield`, including yield earned in an `Anytime` circle. This only becomes correct once yield is circle-scoped.

src/contracts/SavingCircles.sol
```diff
@@ -153,7 +153,7 @@ contract SavingCircles is ISavingCircles, ReentrancyGuardUpgradeable, OwnableUpg
   }
 
   /// @inheritdoc ISavingCircles
-  function decommission(uint256 _id) external override nonReentrant onlyActive(_id) onlyMember(_id, msg.sender) {
+  function decommission(uint256 _id) external virtual override nonReentrant onlyActive(_id) onlyMember(_id, msg.sender) {
```
ss251: **10. Making `decommission` virtual is not enough.** `YieldSavingCircles` never overrides it.

The inherited body `safeTransfer`s `circle.token` amounts from `roundDeposits` out of *this* contract. After `_deposit` those tokens are in CrowdStake (and after the BREAD adapter they have been burned). Members get nothing / the call reverts.

Override must: settle accounting, redeem the correct principal from the module, realize owed yield, convert back to BREAD, then pay members. Also broken further if `_withdraw` zeros `roundDeposits` (point 6).

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+  error YieldModuleNotSet();
+  error YieldLocked();
+  error NothingToClaim();
+
+  /// @notice Set the yield module. Owner only. Must be set before any deposits.
+  function setYieldModule(address _yieldModule) external onlyOwner {
+    if (_yieldModule == address(0)) revert YieldModuleNotSet();
+    yieldModule = IYieldModule(_yieldModule);
+    emit YieldModuleSet(_yieldModule);
+  }
```
ss251: **11. "Must be set before any deposits" is only a comment.** Owner can `setYieldModule` while `_totalPrincipal`, `_principalOf`, `_rewardDebt`, and `_lastYieldSeen` still describe the old position.

Lock after init, or revert unless `_totalPrincipal == 0` (and settle/unwind first if you ever allow rotation). Same class of bug on `setSoleToken` (L274): no relationship to `circle.token` / `allowedTokens`, and it can change while circles exist.

```suggestion
  /// @notice Set the yield module. Owner only. Immutable once any principal is staked.
  function setYieldModule(address _yieldModule) external onlyOwner {
    if (_yieldModule == address(0)) revert YieldModuleNotSet();
    if (_totalPrincipal != 0) revert YieldModuleNotSet();
    yieldModule = IYieldModule(_yieldModule);
    emit YieldModuleSet(_yieldModule);
  }
```

(`YieldModuleNotSet` is reused only as a stand-in — add a dedicated error if you keep a setter at all.)

src/contracts/YieldSavingCircles.sol
```diff
@@ -0,0 +1,277 @@
...
+    return IERC20(_soleToken);
+  }
+
+  /// @dev v1 single-token constraint: set on first allowed token. See #189 (multi-token later).
+  address private _soleToken;
+
+  /// @notice Owner sets the single supported ERC20 for circles + payouts (v1 constraint).
+  function setSoleToken(address _token) external onlyOwner {
+    _soleToken = _token;
+  }
```
ss251: **11. `setSoleToken` has no invariant.** Inherited `create()` still accepts any `allowedTokens` entry. Nothing requires `circle.token == _soleToken`. Owner can retarget payouts while circles are live. Make this immutable at initialize, or require no active circles / no principal, and reject `create` when `token != _soleToken`.

ss251 (changes requested):
Requesting changes before this contract/API becomes something the Stacks UI depends on.

I reviewed PR #190 against issue #189, the current `SavingCircles` state machine, `AutomaticSavingCircles`, and the current CrowdStake token/pool implementations. There are several blockers that need to be resolved at the contract-design level, not just patched around in the frontend.

Inline comments below map to each numbered point. I only attached GitHub suggested-changes where a small, correct edit exists. The rest are design blockers, not one-line patches.

**1. CrowdStake interface drift currently makes deposits incompatible with the concrete implementations.**

`IYieldModule` declares:

```solidity
mint(uint256 amount, address receiver)
```

and this PR calls:

```solidity
yieldModule.mint(_underlyingAmount, address(this));
```

but the current CrowdStake token/pool surfaces implement:

```solidity
mint(address receiver, uint256 amount)
```

These are different ABI selectors (`0x94bf804d` vs `0x40c10f19`). The interface in CrowdStake itself appears stale, so this is not solely a saving-circles issue, but #190 cannot integrate with the concrete implementations until the canonical interface is reconciled.

**2. `claimYield()` is using CrowdStake's donated-yield path as though it were the SavingCircles holder's yield.**

Current CrowdStake distinguishes:

* `yieldAccrued()` — donated/distributable yield
* `totalYieldAccrued()` — total vault surplus
* `keptYieldOf(account)` / `claimKeptYield()` — yield retained by an individual holder
* `setYieldSplit()` — controls how much of a holder's yield is kept vs donated

#190 reads the module-wide `yieldAccrued()`, allocates that locally to SavingCircles members, and calls `claimYield()`.

That conflicts with CrowdStake's current economic model. If SavingCircles is meant to own the yield generated by its CrowdStake position, the integration should explicitly establish that ownership — e.g. via the holder/kept-yield path — rather than treating the donated pool as SavingCircles yield.

This matters especially if the CrowdStake module is shared with other holders: their donated yield can enter the numerator of `_harvestGlobal()` while their principal is absent from `_totalPrincipal`.

**3. The current `claimYield()` call is not authorized on a normally wired CrowdStake instance.**

CrowdStake's concrete token and pool implementations restrict `claimYield()` to the configured `yieldClaimer`.

The standard deployment wiring sets the `DistributionManager` as that claimer.

#190 instead has `YieldSavingCircles` call:

```solidity
yieldModule.claimYield(...)
```

There is no deployment/configuration work in this PR that makes `YieldSavingCircles` the authorized claimer. This needs an explicit integration decision rather than assuming direct access to the donated-yield claim path.

**4. Principal/yield asset types do not line up with the abstraction in #190.**

The current sDAI implementations redeem principal through:

```text
sDAI -> WXDAI -> native xDAI
```

and send native xDAI to the `burn()` receiver.

`YieldSavingCircles` calls:

```solidity
yieldModule.burn(_principal, address(this));
```

but has no payable receiver and then proceeds as if it received an ERC-20 that `_fromUnderlying()` can convert and `_payoutToken().safeTransfer()` can send.

Pool-mode yield claims return WXDAI, while token-mode `SexyDaiYield.claimYield()` mints the CrowdStake token. These are three different asset flows:

* principal redemption: native xDAI
* pool-mode yield: WXDAI
* token-mode yield: CrowdStake token

They cannot currently be represented by a single `_fromUnderlying(uint256)` seam.

**5. The BREAD adapter needs to be redesigned, not just filled in.**

Actual BREAD conversion is stateful:

```text
BREAD burn -> native xDAI
native xDAI / WXDAI -> BREAD mint
```

But `_toUnderlying(uint256)` is called both when a real deposit is made and again later when historical `roundDeposits` are processed during withdrawal.

If `_toUnderlying()` performs the actual BREAD burn, the withdrawal path attempts to convert the historical deposit a second time.

We need to separate:

* asset movement/conversion, and
* amount/accounting conversion.

The current hooks conflate them.

**6. `_withdraw()` clears historical `roundDeposits`, which breaks the base SavingCircles state machine.**

#190 does:

```solidity
roundDeposits[_id][_round][_depositor] = 0;
```

when a pot is claimed.

The inherited `_deposit()` and `_isDecommissionable()` logic rely on those historical values to determine whether the previous round was fully funded.

After round 0 is claimed, its deposit records become zero. When round 1 begins, the inherited logic sees round 0 as incomplete and reverts new deposits with `CircleTimedOut()` / makes the circle decommissionable.

`roundDeposits` is protocol history and cannot be consumed as the yield-principal ledger. Yield redemption state needs separate storage.

**7. `withdrawStakedPrincipal()` withdraws all of a user's principal, not the claimant pot they elected to keep staked.**

Every normal deposit increments:

```solidity
_principalOf[_member]
```

and `withdrawStakedPrincipal()` does:

```solidity
uint256 _principal = _principalOf[msg.sender];
_principalOf[msg.sender] = 0;
yieldModule.burn(_principal, ...);
```

There is no circle id and no separate `keptPotPrincipal`.

A member can therefore withdraw ordinary contributions that are still backing active rounds. `roundDeposits` continues to say those contributions exist, and the later pot claim attempts to subtract that principal again.

This needs separate per-circle balances for mandatory deposits versus a claimant's voluntarily retained pot.

**8. The claimant does not currently control the advertised keep-staked decision.**

The base contract intentionally allows anybody to call:

```solidity
withdrawFor(id, member)
```

for a valid claimant.

The new keep-staked behavior is controlled by a temporary `_keepStakedFlag` that is only set by:

```solidity
claim(id, keepStaked)
```

So another caller — including the existing automation path — can call `withdrawFor(id, claimant)` with `_keepStakedFlag == false` and force the claimant's pot out of the yield position.

The recipient still receives their funds, but their advertised "take vs keep staked" choice has been made by somebody else.

`AutomaticSavingCircles` currently still uses `withdrawFor()`.

**9. The global reward index can be valid, but member state cannot be global if claim policy is per circle.**

A shared `_accYieldPerPrincipal` is reasonable if all circles intentionally share one yield source.

The problematic part is:

```solidity
_principalOf[address]
_rewardDebt[address]
claimableYield[address]
```

A member can participate in multiple circles, while claim timing is configured per circle.

Once yield is collapsed into:

```solidity
claimableYield[Alice]
```

the contract no longer knows which portion belongs to which circle.

That directly breaks `ClaimTiming`. `_yieldUnlocked()` currently blocks all of a member's yield whenever any active circle they belong to is `EndOnly`, including yield earned in an `Anytime` circle.

At minimum, economic ownership needs circle-scoped state, e.g.:

```solidity
principal[circleId][member]
rewardDebt[circleId][member]
claimableYield[circleId][member]
keptPotPrincipal[circleId][member]
```

The reward source/index may remain shared if that is the intended architecture.

**10. `decommission()` is incompatible with yield-bearing custody.**

The inherited implementation returns the ERC-20 amounts recorded in `roundDeposits` directly from `SavingCircles`.

After #190's deposit path, that principal has been moved into CrowdStake.

After the intended BREAD conversion is added, the deposited BREAD will have been burned entirely.

Decommission therefore needs a yield-aware unwind that:

1. settles accounting,
2. redeems the correct principal,
3. realizes/assigns owed yield,
4. converts assets back to BREAD,
5. returns the correct amounts to members.

The current inherited implementation cannot do that.

**11. Important configuration invariants exist only in comments.**

`setYieldModule()` says the module must be set before deposits, but the contract does not enforce that. The owner can change the module while `_totalPrincipal`, `_principalOf`, `_rewardDebt`, and `_lastYieldSeen` still refer to the old position.

Likewise `setSoleToken()` can be changed while circles exist, while inherited `create()` still accepts any token in `allowedTokens`. There is no enforced relationship between `circle.token` and `_soleToken`.

These should be immutable after initialization or guarded by explicit no-active/no-principal invariants.

**12. The UI API is not ready to depend on `claimableYield(address)` as currently defined.**

`claimableYield` is stored/settled state. Fresh sDAI yield is incorporated only when `_harvestGlobal()` / `_settle()` execute.

The UI currently reads the stored mapping and treats zero as "nothing to claim", disabling the claim action. That can report zero while unsettled yield exists.

The contract should expose a current view such as `pendingYield(circleId, member)` / `claimableYield(circleId, member)` that simulates the accumulator without requiring a state-changing transaction.

---

### Required direction before this API stabilizes

Before we extend the UI around this ABI, I think the contract work needs to establish these invariants explicitly:

1. Decide whether SavingCircles gets a dedicated CrowdStake instance or uses holder-level isolation inside a shared one.
2. Use the canonical CrowdStake deposit interface and asset semantics.
3. Define how SavingCircles legally/economically owns and claims its yield (`keptYield` path vs a dedicated claimer architecture).
4. Separate BREAD conversion from accounting-only amount conversion.
5. Preserve `roundDeposits` as immutable historical ROSCA bookkeeping.
6. Track principal, reward debt, claimable yield, and retained claimant pots by circle/member.
7. Make `withdrawStakedPrincipal` circle-scoped and limited to retained pot principal.
8. Resolve `withdrawFor` / automation semantics for the claimant's keep-staked choice.
9. Implement yield-aware decommissioning.
10. Lock or validate the yield module/token configuration once funds exist.
11. Add a current read-only yield calculation for the UI.
12. Add tests covering at least:

* multiple members depositing at different times,
* round 0 claim followed by successful round 1 deposits,
* one member participating in multiple circles,
* Anytime + EndOnly circles simultaneously,
* keep-staked pot followed by later mandatory deposits,
* unauthorized principal withdrawal attempts,
* decommission with principal + accrued yield,
* BREAD -> xDAI/WXDAI -> sDAI -> BREAD round trip,
* CrowdStake claim authorization / holder yield ownership,
* AutomaticSavingCircles behavior,
* proxy/storage upgrade safety.

One CI note: the current GitHub `Build & Test` failure did **not** reach Solidity compilation. Dependency installation failed while fetching `forge-std`, so `forge build` and tests were skipped. Compilation therefore remains unverified rather than known-failing.

Given the state-machine issues and the unresolved CrowdStake/BREAD ownership model, I'm requesting changes rather than treating these as follow-up cleanup. The UI PR should remain decoupled from this ABI until these contract invariants are settled and covered by integration tests.

src/contracts/SavingCircles.sol
ss251: **Amendment (author DMs):** decommission is not "return each member their yield." At-fault members' *unclaimed* yield is meant to be spread to not-at-fault members as a reward for staying current.

That is a scope change vs #189 / this PR's "no forfeiture" language, and it cannot be implemented on the inherited body or on global `claimableYield[address]`. It also depends on `roundDeposits` remaining intact so "at fault" is computable.

Open: exact at-fault definition, yield-only vs principal, no clawback of Anytime claims, and what happens to keep-staked pots. Full writeup in the conversation amendment.

ss251:
## Amendment to the review above — decommission yield forfeiture

New author input (DMs), not in #189 or this PR:

> If a circle is decommissioned then the yield accumulated for the users at fault will be spread to the users who are not at fault. So it's more like a reward for completing the circle.

That **does** change the review. Specifically point 10, and it makes point 9 / point 6 blocking rather than cleanup.

### What it contradicts

- #189: **No forfeiture** — defaulters are not slashed (explicitly out of scope for v1).
- This PR's header: "No forfeiture (out of scope for v1)."
- #189 acceptance: "Decommission returns principal + owed yield correctly" — which reads as each member keeping *their own* yield.

This is yield forfeiture / redistribution, not principal slashing. It needs to be written into #189 as a v1 scope change before this ABI is treated as settled.

### What a correct unwind becomes

Not "return everyone their principal + their yield." Roughly:

1. Settle per-circle / per-member accounting
2. Classify at-fault vs not-at-fault
3. Redeem remaining principal from CrowdStake
4. Forfeit **unclaimed** yield of at-fault members and add it to not-at-fault members
5. Convert back to BREAD
6. Return each member their remaining principal + (possibly redistributed) yield

The inherited `decommission()` still `safeTransfer`s `circle.token` out of this contract from `roundDeposits` and cannot do any of that.

### Why the rest of the review gets stricter

- **Point 6:** `roundDeposits` is now also the only honest "who defaulted" signal. Zeroing it on claim makes this rule uncomputable.
- **Point 9:** `claimableYield[alice]` cannot forfeit "Alice's yield *in this circle*." Circle-scoped yield is required for this, not just for `ClaimTiming`.
- **Point 12:** after decommission the UI must show *redistributed* yield, not the pre-forfeiture accrual.

### Open questions the contract cannot invent

- **Who is at fault?** Natural reading of today's trigger: at fault iff `roundDeposits[id][failedRound][member] < depositAmount`. Confirm whether already-claimed members or partial deposits change that.
- **Principal is not slashed?** Today's `decommission()` returns each depositor their own remaining principal. Confirm yield-only.
- **No clawback of already-claimed yield** (`ClaimTiming.Anytime`)? Only unclaimed yield can be redistributed.
- **Keep-staked pots:** if a not-at-fault claimant left their pot staked, is that pot + later yield theirs, or in the redistribution pool?
- **"Reward for completing the circle"** is slightly the wrong phrase — decommission means the circle *did not* complete. The reward is for *not defaulting*. Worth naming that in #189 so the UI does not promise a completion bonus.

Until those are specified in #189 and implemented with tests (including Anytime + keep-staked + multi-circle), this stays a request-changes blocker, not follow-up cleanup.

Issue #131, opened 2026-03-11
General message to retire the Stack for users even tho they weren't part of the Stack


brendaasfjnso:
<img width="646" height="362" alt="Image" src="https://github.com/user-attachments/assets/29a579d3-576c-4023-9185-edb50b1c48a2" />

<img width="659" height="381" alt="Image" src="https://github.com/user-attachments/assets/b215f4d1-b5e9-4b41-aaa3-f54ad066663e" />

<img width="889" height="424" alt="Image" src="https://github.com/user-attachments/assets/ade00578-3bf5-477f-9381-abdb87877ceb" />

"Seems there is an error making a tester not able to access their group"

exo404:
I'm fixing this from the contract by giving permission to decommission only for the stack members, but I think we need some changes on the UI too
@Olaleye-Blessing @brendaasfjnso @diterra-code

PR #192, opened 2026-08-23
Feat/migrate member

Olaleye-Blessing:


franrolotti:
Had a look — the approach makes sense to me. Replacing in place to preserve
`memberIndex` seems like the right call, and it's the only path I can see that works
post-start. A few thoughts, mostly questions:

---

One thing I'd like to make sure doesn't get lost: this migrates *accounting only* —
there's no token transfer in `migrateMember`. So whatever sits in the embedded wallet
itself stays there. I've got ~70 BREAD in mine that isn't in any stack, and I'd guess
that's fairly common. It'd be easy for someone to complete the migration, see it
succeed, and not realize their balance didn't come along.

I know the sweep is planned frontend-side — just flagging two things that might be
worth building into it:

- **Sweeping native gas as well as BREAD.** If I'm reading app-stacks #131 right,
  sponsorship becomes embedded-only, so an EOA with no gas can't sign anything —
  including the approvals below. Might be worth confirming.
- **Making it resumable.** It's several transactions across two wallets. If someone
  closes the tab midway, could they end up in a partial state that's hard to detect?

Rough order that seems safest to me: sweep funds → EOA approves both contracts →
`migrateMember` → `migrateAutomationSettings`. Happy to be wrong on the ordering.

---

**Tests** — this rewrites a fair bit of state on a live circle. Could we add coverage
for mid-circle migration preserving payout position, `roundDeposits` carry-over across
rounds, and migrating after `hasClaimed` is true?

**`migrateAutomationSettings` and allowances** — automated deposits do
`safeTransferFrom(_member, ...)` and check `allowance(_member, address(this))`, and
allowances don't migrate. So once the flags point at the new address, if it has no
allowance or no balance the automated deposit would revert and the member misses the
round — which I think can make the circle decommissionable for everyone, not just
them. Would it be worth reverting unless `isMember(circleId, newMember)`, or at least
noting in the NatSpec that the new address needs to approve first? Right now there's
no membership check.

**Ordering between the two contracts** — they're separate txs on separate contracts,
so if the second doesn't land the flags sit on an address that's no longer a member.
Probably fine if the frontend guarantees ordering, but maybe worth a NatSpec line
saying they're a pair?

**Scope** — framed as wallet migration, but it's a general mid-circle transfer of a
payout slot to any address. Is that intentional? Also curious how this sits alongside
#148 (binding invites to a recipient address) — they seem to approach the same
wrong-address problem from opposite directions, so maybe worth deciding which one we
want.

**Nit** — `balances` is declared `mapping(uint256 id => mapping(address token => uint256 balance))`
but every write keys it by member. The `token` label had me misreading lines 316-317
as dead code at first. Pre-existing, not this PR, but a rename might save someone else
the confusion.

